PR #12, opened 2015-01-16
ax_compiler_flags: Add two extra warning CFLAGS

pwithnall:
-Wimplicit-function-declaration and -Wreturn-type. These are implicitly
activated by -Wall on GCC, but might not be on other compilers. Best to
be explicit.


peti:
Merged in db885c0. Thank you for the patch.

PR #13, opened 2015-01-26
Split AX_COMPILER_FLAGS into sub-macros for different tools

pwithnall:
See the two commit messages for justification. Split AX_COMPILER_FLAGS (without breaking its API) into sub-macros for different tools, initially supporting the C compiler and linker, and then adding support for g-ir-scanner from GObject Introspection (https://wiki.gnome.org/Projects/GObjectIntrospection).


peti:
Merged, thank you very much for the update.

One minor nit:: if you have the chance, please run `make maintainer-all all` before pushing to check that the format stuff is in order, i.e. `#serial` numbers are bumped, etc. Thank!

PR #15, opened 2015-01-30
Test that we can actually #include <omp.h> successfully

jwpeterson:


peti:
Could you please force-push a new version that bumps the `#serial` number, too, to make the Travis-CI job succeed?

PR #19, opened 2015-02-05
ax_compiler_flags: Add C++ compiler warning checks

amigadave:
In EasyTAG, I use C and C++ within the same code base, and it helps to distinguish between the (quite overlapping) sets of warning flags for each compiler. Add a new AX_COMPILER_FLAGS_CXXFLAGS macro, based on AX_COMPILER_FLAGS_CFLAGS, and use it within AX_COMPILER_FLAGS to AC_SUBST the available warnings as WARN_CFLAGS. Take extra warnings from the supplied EXTRA-*-CFLAGS variables, which may or may not be the best approach, as most warnings are common between the C and C++ compilers, but some are not.

### DIFF
--- a/m4/ax_compiler_flags.m4
+++ b/m4/ax_compiler_flags.m4
@@ -55,6 +55,11 @@
 #   be manually added to the CFLAGS and LDFLAGS variables for each target in
 #   the code base.
 #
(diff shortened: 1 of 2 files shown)


peti:
Looks fine to me. @pwithnall, do you have any comments about this change?

pwithnall:
All looks good to me.

PR #21, opened 2015-02-11
Added macro to detect Haxe <http://haxe.org> version.

Jens-G:
I did not found any, but needed it, so I wrote one. The existing Ruby macro served me as a template.


peti:
Merged in b6111ae. Thank you very much.

PR #23, opened 2015-02-17
bootstrap.sh: Support a --copy option to automatically clone gnulib

pwithnall:
When building autoconf-archive using build bots or JHBuild, it is quite
desirable to have a single command which will bootstrap the entire build
process, up to the point where `make` can be run.

bootstrap.sh is supposed to do this, but will fail if gnulib is not
present. Add a --copy option which will automatically clone gnulib into
the source directory if it’s not present.

### DIFF
--- a/bootstrap.sh
+++ b/bootstrap.sh
@@ -2,6 +2,10 @@
 
 set -eu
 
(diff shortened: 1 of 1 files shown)


peti:
Thank you for the patch; that's a nice idea.

PR #25, opened 2015-02-20
ax_compiler_flags: Conditionally enable C++ warnings

amigadave:
Inflicting the requirement of a C++ compiler on all projects using AX_COMPILER_FLAGS was not intended, and this pull request should (mostly) fix that. I cannot think of a neat way to have this work if AX_COMPILER_FLAGS occurs before AC_PROG_CXX, so the current solution deserves a careful examination, but is no worse than what is in master.

I have tested this with EasyTAG (which has both C and C++ sources) and Yelp (which has only C sources) and it seems to work as expected.

The ax_compiler_flags_cflags change is not related, just a bit of defensive programming.

### DIFF
--- a/m4/ax_compiler_flags.m4
+++ b/m4/ax_compiler_flags.m4
@@ -59,10 +59,11 @@
 #   be manually added to the CFLAGS and LDFLAGS variables for each target in
 #   the code base.
 #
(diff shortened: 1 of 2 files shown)


peti:
Merged in 1125e05 and 9b3173a.

PR #27, opened 2015-02-26
Fix lua prefix path detection

opoplawski:
This fixes lua prefix path detection by unquoting a configure shell variable that needs to be expanded at run time and by passing in the proper arguments.


gvvaughan:
@rrthomas please review when you have time

rrthomas:
Thanks, I pushed this upstream.

PR #28, opened 2015-03-26
AX_PATH_GENERIC fixes for flags containing spaces

wsfulton:
ax_path_generic.m4 fix when the _CFLAGS and _LIBS are provided and they
contain a space

For example:
./configure PCRE_LIBS='-L/home/me/pcre/lib -lpcre'

Patch also contains a portability fix from @vadz:
Use POSIX "=" instead of "==" which is not supported at least by FreeBSD
/bin/sh.

### DIFF
--- a/m4/ax_path_generic.m4
+++ b/m4/ax_path_generic.m4
@@ -69,7 +69,7 @@
 #   and this notice are preserved. This file is offered as-is, without any
 #   warranty.
 
-#serial 11
+#serial 12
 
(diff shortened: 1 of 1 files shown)


wsfulton:
I see, thanks for explanation, all is clear now! I saw the Travis failure beforehand but didn't know what to do as it really isn't obvious why it failed. I've fixed and rebased the pull request and tests are passing now.

PR #29, opened 2015-03-26
AX_BOOST_BASE: Support for native Windows header only libraries

wsfulton:
Native Windows versions of Boost (for Visual C++) have a different
    layout to other platforms. Typically, one has:
    
    <root>\boost\version.hpp
    
    This patch adds support for header only libraries.


wsfulton:
I see, thanks for explanation, all is clear now! I saw the Travis failure beforehand but didn't know what to do as it really isn't obvious why it failed. I've rebased the pull request and tests are passing now.

PR #30, opened 2015-03-28
ax_python_module_version: Add initial macro.

Kurt-von-Laven:
The Python module version macro builds on top of the existing Python
module macro authored by Andrew Collier. The Python module macro checks
that a Python module is installed, and the Python module version macro
tests that a Python module is installed at a given version or higher.

### DIFF
--- a/m4/ax_python_module_version.m4
+++ b/m4/ax_python_module_version.m4
@@ -1,0 +1,35 @@
+# ============================================================================
+#  http://www.gnu.org/software/autoconf-archive/ax_python_module_version.html
+# ============================================================================
+#
+# SYNOPSIS
+#
+#   AX_PYTHON_MODULE_VERSION(modname, min_version)
+#
+# DESCRIPTION
+#
+#   Checks for Python module with at least the given version.
+#
+#   Triggers an error if module is absent or present but at a lower version.
+#
+# LICENSE
+#
+#   Copyright (c) 2015 Endless Mobile, Inc.; contributed by Philip Chimento <philip@endlessm.com> and Kurt von Laven <kurt@endlessm.com>
+#
+#   Copying and distribution of this file, with or without modification, are
+#   permitted in any medium without royalty provided the copyright notice
+#   and this notice are preserved. This file is offered as-is, without any
+#   warranty.
+
+#serial 1
+
+AC_DEFUN([AX_PYTHON_MODULE_VERSION], [
+    AX_PYTHON_MODULE([$1], [required])
+    AC_MSG_CHECKING([for version $2 or higher of $1])
+    $PYTHON -c "import sys, $1; from distutils.version import StrictVersion; sys.exit(StrictVersion($1.__version__) < StrictVersion('$2'))" 2> /dev/null
+    AS_IF([test $? -eq 0], [], [
+        AC_MSG_RESULT([no])
+        AC_MSG_ERROR([You need at least version $2 of the $1 Python module.])
+    ])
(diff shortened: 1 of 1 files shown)


Kurt-von-Laven:
Here is the build error for the sake of convenience:

```
python ./macro2m4.py "m4/ax_python_module_version.m4" "stage/ax_python_module_version.m4"
Traceback (most recent call last):
  File "./macro2m4.py", line 56, in <module>
    m = Macro(m4File, computeSerialNumber=True)
  File "/home/travis/build/peti/autoconf-archive/macro.py", line 105, in __init__
    raise Exception("%s: unknown section %r in macro" % (filePath, key))
Exception: m4/ax_python_module_version.m4: unknown section 'authors' in macro
```

My interpretation of the build error is that the build didn't like the "AUTHORS" section. What is the preferred syntax for multiple authors when the copyright holder is an organization?

peti:
What we usually do is to list both the individuals who wrote the macro as well as the organization as copyright holders in the license section.

m4/ax_python_module_version.m4
```diff
@@ -0,0 +1,38 @@
...
+#
+#   Copying and distribution of this file, with or without modification, are
+#   permitted in any medium without royalty provided the copyright notice
+#   and this notice are preserved. This file is offered as-is, without any
+#   warranty.
+
+AC_DEFUN([AX_PYTHON_MODULE_VERSION], [
+    AX_PYTHON_MODULE([$1], [required])
+    AC_MSG_CHECKING([for version $2 or higher of $1])
+    $PYTHON -c "import sys, $1; from distutils.version import StrictVersion; sys.exit(0 if StrictVersion($1.__version__) >= StrictVersion('$2') else 1)" 2> /dev/null
```
wsfulton: Python conditional expressions were introduced in Python 2.5. As this is a version checking macro, the code should be changed to work with versions of Python older than 2.5!

Kurt-von-Laven: Good point! Just pushed a new version sans the conditional expression.

Kurt-von-Laven:
Thanks. I discussed this with my team, and we aren't supposed to list individuals as copyright holders because of a standard clause in our employment contract. Just pushed a revised version that lists the authors as points of contact beneath the copyright, but open to other suggestions.

peti:
I find it hard to wrap my head around the fact that you guys release a macros under the all-permissive license, which which gives the whole world the right to do with it, basically, whatever they please, yet at the same time you don't give the people who wrote that 9 line macro the right to list their names in the copyright line? That feels weird to me.

Anyway, this new version introduces the AUTHORS section through the back door, only it doesn't call it so. Considering that, I'd rather add a proper AUTHORS section that put free-style comments in the license, i.e. those licenses ought to be machine-readable with relatively minor effort.

How about the following compromise: In `ax_cxx_compile_stdcxx_11.m4` the submitter used the copyright line:

> Copyright (c) 2014, 2015 Google Inc.; contributed by Alexey Sokolov <sokolov@google.com>

Would that style be okay for you?

Kurt-von-Laven:
It's not like that; I just don't think it's worth heckling our CEO and lawyers to put together an exemption to the pertinent clause in our employment contract for the sake of this macro. The style you described works for us, and I pushed a revised version.

Kurt-von-Laven:
Is there anything else that I should change here?

peti:
Sorry, I forgot to merge this patch! It's done now. Thank you for your efforts!

Kurt-von-Laven:
No worries; thank you!

PR #31, opened 2015-04-06
ax_lib_postgresql: Add convenience _LIBS variable and fix version checking

totakura:
This pull request adds a new _LIBS variable to used as a convenience variable.  It also fixes version checking outcome; earlier, HAVE_POSTGRESQL is set when given optional version is not present.


totakura:
Wonder why the build fails; there seems to be a problem with serial bumping.

totakura:
A better version of this pull request is available at https://github.com/peti/autoconf-archive/pull/32.  Closing this one.

PR #33, opened 2015-04-16
Look for make using the current environment instead of spawning a new shell and extract version

emcrisostomo:
Hi.

The current implementation searches for `make` executing a command using `sh -c`, so that current environment variables are not considered. Additionally, the version is extracted and saved into a variable.

Regards,

Enrico

### DIFF
--- a/m4/ax_check_gnu_make.m4
+++ b/m4/ax_check_gnu_make.m4
@@ -8,12 +8,18 @@
 #
 # DESCRIPTION
 #
-#   This macro searches for a GNU version of make. If a match is found, the
-#   makefile variable `ifGNUmake' is set to the empty string, otherwise it
-#   is set to "#". This is useful for including a special features in a
-#   Makefile, which cannot be handled by other versions of make. The
-#   variable _cv_gnu_make_command is set to the command to invoke GNU make
-#   if it exists, the empty string otherwise.
+#   This macro searches for a GNU version of make. If a match is found:
+#
+#     * The makefile variable `ifGNUmake' is set to the empty string, otherwise
+#       it is set to "#". This is useful for including a special features in a
+#       Makefile, which cannot be handled by other versions of make.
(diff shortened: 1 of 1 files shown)


emcrisostomo:
Hi @peti, I hope you don't mind if I ask what is going on with this build. After the first error, where the details showed a diff where the file generated by `macro2m4.py` had a newline into the header comment, I added a commit to this PR with that change. Now I see that the build fails again and the difference is the serial number:

```
% python ./macro2m4.py m4/ax_check_gnu_make.m4 stage/ax_check_gnu_make.m4                                        git:feature/ax_check_gnu_make
% diff m4/ax_check_gnu_make.m4 stage/ax_check_gnu_make.m4                                                        git:feature/ax_check_gnu_make
64c64
< #serial 8
---
> #serial 9
```

This macro's previous serial number was 7, and I bumped it to 8. What am I doing wrong?

Thanks,

Enrico

peti:
Hi @emcrisostomo, the `#serial` number is the number of commits that modify the macro, so when you added a commit to bump `#serial`, it required another bump because of the additional commit. :grin:

The easiest way to remedy this issue is to run `git rebase -i origin/master`, to squash both commits into one, and then to do a force push: `git push -f`. That should fix the travis-ci build.

emcrisostomo:
Hi @peti, thanks for the clarification. I'm closing this PR and opening another one.

peti:
Why did you close the PR? The "force push" exists precisely so that you can make those kind of re-arrangements with an existing PR.

emcrisostomo:
You are right @peti, but I must have done something wrong because (as you can see above) I ended up increasing the list of commits in this PR rather than squashing it into one. I thought at this moment it was just easier for me to re-create it than fiddling with this one. I guess this Mercurial man should practice his git-foo. ;)

peti:
OK. :grin:

PR #34, opened 2015-04-17
Look for make using the current environment instead of spawning a new shell

emcrisostomo:
Look for make using the current environment instead of spawning a new shell.
Extract the GNU Make version and store it into a variable.


emcrisostomo:
Thanks for your assistance @peti.

peti:
You're welcome. Thank you for the patch!

PR #35, opened 2015-05-02
ax_python_module[_version]: Add Python 3 support.

Kurt-von-Laven:
Permit dependency checking of both Python 2 and Python 3 modules.

### DIFF
--- a/m4/ax_python_module.m4
+++ b/m4/ax_python_module.m4
@@ -4,13 +4,15 @@
 #
 # SYNOPSIS
 #
-#   AX_PYTHON_MODULE(modname[, fatal])
+#   AX_PYTHON_MODULE(modname[, fatal, python])
 #
 # DESCRIPTION
 #
 #   Checks for Python module.
 #
 #   If fatal is non-empty then absence of a module will trigger an error.
+#   The third parameter can either be "python" for Python 2 or "python3" for
+#   Python 3; defaults to Python 3.
 #
 # LICENSE
 #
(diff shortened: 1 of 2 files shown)


m4/ax_python_module.m4
```diff
@@ -21,13 +23,13 @@
...
 
-#serial 6
+#serial 7
 
 AU_ALIAS([AC_PYTHON_MODULE], [AX_PYTHON_MODULE])
 AC_DEFUN([AX_PYTHON_MODULE],[
     if test -z $PYTHON;
     then
-        PYTHON="python"
+        PYTHON="$3"
```
peti: What happens here when `$3` is unset?

Kurt-von-Laven: Currently that will confuse this code into thinking that the specified Python module can't be found. Would you prefer that we give a more helpful error message in this case saying too few parameters were passed? Alternatively, we could default the third parameter to "python."

peti: Well, the synopsis suggests that the parameter is optional, so IMHO the macros should do something useful even if the 3rd parameter is left unspecified.

Kurt-von-Laven: Doh; I didn't mean to mark the parameter optional, but it seems like a good idea to me. Now the third parameter defaults to "python3," but I'm definitely open to other suggestions.

Kurt-von-Laven:
I see that GitHub hid my reply, so just wanted to let you know that I believe I have addressed your comment.

peti:
Very cool. Thank you for your efforts!

Kurt-von-Laven:
Thank you!

PR #37, opened 2015-06-14
ax_compiler_flags: Require AX_COMPILER_FLAGS_GIR to be defined

phako:
Be consistent. _C(XX)FLAGS and _LDFLAGS are required as well and gives an easier to understand  error during autoreconf


peti:
Can you please amend this commit with a bump of the `#serial` number? The changes should pass `make maintainer-all all`, and currently they don't.

phako:
Oh, sorry. Will do.

peti:
Thank you very much for the update!

PR #38, opened 2015-07-02
Add GNU Bison version check

PSRCode:
Valid from Bison v1.29b until now[1].

[1]http://git.savannah.gnu.org/cgit/bison.git/commit/?id=e79137accc7ea0352cd4677ff22818f9c68d4eab

### DIFF
--- a/m4/ax_prog_bison_version.m4
+++ b/m4/ax_prog_bison_version.m4
@@ -1,0 +1,68 @@
+# ===========================================================================
+#   http://www.gnu.org/software/autoconf-archive/ax_prog_bison_version.html
+# ===========================================================================
+#
+# SYNOPSIS
+#
+#   AX_PROG_BISON_VERSION([VERSION],[ACTION-IF-TRUE],[ACTION-IF-FALSE])
+#
+# DESCRIPTION
+#
+#   Makes sure that bison version is greater or equal to the version indicated.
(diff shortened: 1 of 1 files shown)


peti:
IMHO, it would be much nicer to merge that functionality into the existing `ax_prog_bison` macro.

PSRCode:
I did think of doing this but I saw that perl had a basic version check so i was'nt sure. Anyway.
TBH I'm not that familiar on how m4 marcro work so what would the best way of merging it ? Adding a third parameter for version check?

peti:
Hmm, on second thought, you might be right. I guess I'm just worried about duplicating functionality in both `ax_prog_bison` and `ax_prog_bison_version`. It would be nice of the latter were built on top of the former in a way that re-uses the older macro's intelligence.

PSRCode:
As in do the job of the older one and check for version? or as in use the info cached from the older one? In the second case there is not much to use except the cached ax_cv_prog_bison. I'm not a fan on duplicating stuff either but there is not much to reuse.

We could cache the output from "$BISON --version" than use it inside ax_prog_bison_check but TBH this is even more complicated than only 2 files doing each their respective task. I'm not sure where to go from there.

On another subject should we cache the version number ?

peti:
I think you are right. Let's just add this macro in addition to the existing one, and then we can replace the old macro with the new one at a later point.

PSRCode:
Thx

PR #40, opened 2015-07-06
ax_append_flag: Fix regression if flag has a ','.

Batchyx:
Below is a small fix for AX_APPEND_FLAG when the flag to be appended has a comma in it.


peti:
Merged in c9add3c.

PR #41, opened 2015-07-08
ax_valgrind_check: Fix a bash-4-ism

pwithnall:
The ‘2&>’ syntax requires a recent version of bash. Use ‘>… 2>&1’
instead to reduce the requirement.


peti:
Merged in 40f2088.

PR #42, opened 2015-08-07
Check for attribute support with C++11 macro

pnorman:
Some compilers pass the other tests while not implementing this part of the C++11 specification. The test used also does not cause warnings with GCC or Clang, and can be used with -Werror.

After this change, GCC 4.7 no longer passes the tests. This macro is about compiler support, not standard library support, but the libstdc++ that comes with GCC 4.7 doesn't support emplace on standard library containers, so many C++11 programs would fail at compile-time on most systems using GCC 4.7.

This is my first autoconf archive submission, so I'm not sure if there's anything else I need to add.

Cross-ref openstreetmap/osm2pgsql#434


pnorman:
MSVC 2013 and earlier will fail this check, in spite of having good overall C++11 support. The problem is each compiler implements a subset of the standards, and they're all different. What is needed is a compiler which supports the subset of C++11 that the **user** needs.

[MSVC 2015 supports attributes](http://blogs.msdn.com/b/vcblog/archive/2015/06/19/c-11-14-17-features-in-vs-2015-rtm.aspx)

cc @joto

PR #48, opened 2015-11-23
Patches #8766 and #8772

olaf-mandel:
This pull-request contains the following savannah patches:

* [#8766: AX_CODE_COVERAGE: further silence rules for V=0](https://savannah.gnu.org/patch/?8766)
* [#8772: AX_CODE_COVERAGE: add CPPFLAGS to skip assertions](https://savannah.gnu.org/patch/?8772)


peti:
Pushed upstream. Thank you very much for the patches!

PR #53, opened 2016-01-02
Fix ax_boost_system for new versions of boost.

bogado:
Newer version of boost::system have a breaking change that makes
the test on ax_boost_system fail even though the library is
present.

Changed the test to use another construct that did not change. Also
reseted the "CXX_FLAG" to an empty value in case of it having a
"-Werror" that also makes the test break.

----

this update "works on my box" but I have limited ability to test on older versions of boost or with different flags combinations.

### DIFF
--- a/m4/ax_boost_system.m4
+++ b/m4/ax_boost_system.m4
@@ -31,7 +31,7 @@
 #   and this notice are preserved. This file is offered as-is, without any
 #   warranty.
 
-#serial 17
+#serial 18
 
 AC_DEFUN([AX_BOOST_SYSTEM],
 [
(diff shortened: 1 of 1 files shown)


peti:
Please don't forget to bump the `#serial` number which must be identical to the number of commits that modified this particular file. Otherwise, the regression test suite at travis-ci won't succeed: https://travis-ci.org/peti/autoconf-archive/builds/99844813#L456

bogado:
Sorry about that, I might have not read that I had to this.

Just out of curiosity, why add this manual step?

peti:
> Why add this manual step?

Is there any alternative to a manual step? What else would you suggest we do?

PR #57, opened 2016-01-31
Use AM_DISTCHECK_CONFIGURE_FLAGS instead of DISTCHECK_CONFIGURE_FLAGS

pwithnall:
DISTCHECK_CONFIGURE_FLAGS (no prefix) is for the user to override the
distcheck configure flags. The AM_-prefixed version is for build systems
to use.

https://www.gnu.org/software/automake/manual/html_node/Checking-the-Distribution.html

### DIFF
--- a/m4/ax_code_coverage.m4
+++ b/m4/ax_code_coverage.m4
@@ -49,7 +49,7 @@
 #
 # LICENSE
 #
-#   Copyright (c) 2012 Philip Withnall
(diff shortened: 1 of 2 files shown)


peti:
Could you please force-push an update that bump the macro's `#serial` numbers to fix the unit tests at https://travis-ci.org/peti/autoconf-archive/builds/106027596?

pwithnall:
Sorry, I forget that every time. Done.

PR #65, opened 2016-03-09
Improve OpenGL checks

dryman:
Added hooks for action-if-found / action-if-not-found. Also fix many OSX compatibility problems.
Has been tested on OSX and Linux.

### DIFF
--- a/m4/ax_check_gl.m4
+++ b/m4/ax_check_gl.m4
@@ -193,11 +193,6 @@ AC_DEFUN_ONCE([_WITH_XQUARTZ_GL],[
 ])
 
 AC_DEFUN([_AX_CHECK_DARWIN_GL], [ 
-dnl TODO I think I can switch back to with-x
-dnl opengl -> GL_CFLAGS += X_CFLAGS 
-dnl           GL_LIBS += X_LIBS
-dnl glut   -> use header GL/glut.h
-dnl        -> use search lib -lglut
  AC_REQUIRE([_WITH_XQUARTZ_GL])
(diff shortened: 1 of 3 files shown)


peti:
Would it be possible to factor the definitions of `AX_SAVE_FLAGS_WITH_PREFIX` etc. out into separate archive entries that can be re-used by other macros? We do have a bit of a convention that each file defines only one macro, and it would be great to stick to that rule while also improving the visibility of the macros you've added.

Also, I wonder whether you could please run `make maintainer-all all` locally before pushing your changes to ensure that the texinfo renderer is happy with the formatting of the m4 files?

peti:
Pinging @bastien-roucaries for comments.

dryman:
Sure. Any other checks I should perform before resubmission? Thanks.

dryman:
Closing this PR. Want to rebase the commit so the ChangeLog is cleaner.

bastien-roucaries:
Le 9 mars 2016 16:38:29 GMT+01:00, Felix Chern <notifications@github.com> a écrit :
>Sure. Any other checks I should perform before resubmission? Thanks.


 I think it need major refactoring....

Framework was needed on old mac osx....

Ldhack also...

I will get a glimpse



>
>---
>Reply to this email directly or view it on GitHub:
>https://github.com/peti/autoconf-archive/pull/65#issuecomment-194350478

-- 
Envoyé de mon appareil Android avec K-9 Mail. Veuillez excuser ma brièveté.

PR #66, opened 2016-03-09
Improved three opengl autoconf macros

dryman:
Main features added:
- Support action-if-found and not-found hook
- Export HAVE_GL, HAVE_GLU, and HAVE_GLUT preprocessor symbol
- Add necessary flags to CFLAGS/LIBS by default, but can be disabled by
  overwritting action-if-found hook.
- Fail the configure script if lib not found. This feature can be disabled by
  overwritting action-if-not-found hook
- Fix problems on Mac OSX. Been tested on Linux (Ubuntu) and OSX
- Use --with-xquartz-gl instead of --with-gl=x for using X11 on OSX

Two new macros added:
- ax_save_flags_with_prefix.m4: save flags and pushes flag with prefix to the
  flag.
  AX_SAVE_FLAGS_WITH_PREFIX([GL],[[CFLAGS],[LIBS]]) would expands to
    gl_saved_flag_cflags="$CFLAGS"
    gl_saved_flag_libs="$LIBS"
    CFLAGS="$GL_CFLAGS $CFLAGS"
    LIBS="$GL_LIBS $LIBS"
- ax_restore_flags_with_prefix.m4: restore the flags.

### DIFF
--- a/m4/ax_check_gl.m4
+++ b/m4/ax_check_gl.m4
@@ -4,29 +4,36 @@
 #
 # SYNOPSIS
 #
-#   AX_CHECK_GL
+#   AX_CHECK_GL([ACTION-IF-FOUND],[ACTION-IF-NOT-FOUND])
 #
 # DESCRIPTION
 #
-#   Check for an OpenGL implementation. If GL is found, the required
-#   compiler and linker flags are included in the output variables
-#   "GL_CFLAGS", "GL_LIBS", "GL_LDFLAGS", respectively. If no usable GL
-#   implementation is found, "no_gl" is set to "yes".
+#   Checks for an OpenGL implementation. If a valid OpenGL implementation is
+#   found, this macro would set C preprocessor symbol HAVE_GL to 1.
 #
-#   You could disable OpenGL using --with-gl=no
+#   If either a valid OpenGL header or library was not found, by default the
+#   configuration would exits on error. This behavior can be overwritten by
+#   providing a custom "ACTION-IF-NOT-FOUND" hook.
 #
-#   You could choose a specific OpenGL libs using --with-gl=lib_name
+#   If the header, library was found, and been tested for compiling and
+#   linking the configuration would export the required compiler flags to
+#   "GL_CFLAGS" and "GL_LIBS". These two variables can also be overwritten
+#   by user from the command line if they want to link against the library
+#   they specified instead of having the configuration script to detect the
+#   flags automatically. Note that having "GL_CFLAGS" or "GL_LIBS" set
+#   doesn't mean it can compile or link with the flags, since it could be
+#   overwritten by user. However the "HAVE_GL" symbol and "ACTION-IF-FOUND"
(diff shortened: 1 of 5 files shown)


peti:
Very nice, thank you for the patch!

Cc: @bastien-roucaries

bastien-roucaries:
Could I ask a revert ?

I think it could be done atomically and not a a huge patch.

Some hack were needed

peti:
@bastien-roucaries, why don't you simply submit a PR with an improved version? I'm not sure I see a technical reason for reverting those changes. Do they break anything?

bastien-roucaries:
Yes,

First change are not atomic and reviewable. If it break we could not revert
back.

Moreover some mac osx hack are needed on old version.

And I do not have time to get detail review (I suspect it could break a few
thing on chrome)

Bastien

dryman:
I submit this PR because the previous version doesn't work on my Mac. I looked into the source code and found some bugs that were hard to fix. For example line 215 in `ax_check_gl.m4` saved flags twice and will lose track of flags that has been saved. There was another misquote causes the configure script to look at a macro that should be expanded but didn't:

```
./configure: line 6787: _AX_CHECK_GL_MANUAL_HEADERS_DARWIN: command not found
```

I dropped the support of general darwin, because I couldn't even spin up a darwin VM, either through virtual-box or qemu (the image was downloaded from darwin official website, but still). Thus I made the decision that the macro only need to consider OSX that has OpenGL framework. This simplifies the support on Mac OSX. I don't have machines to test on very old OSX platform that doesn't have OpenGL framework (earlier than 10.4? that's more than ten years) but all the OSX systems that has OpenGL should work with the `-framework OpenGL` option.

I'm more than happy to help to improve the OpenGL checks. If you found any project that breaks with this PR please let me know. The PR I submitted follow many autoconf coding conventions. I hope this can help the review and future maintenance.

dryman:
I searched a bit on how chrome is using ax_check_gl.m4. 
Seems that there was a [PR to google chromium project](https://bugs.chromium.org/p/webp/issues/detail?id=211) that mentioned ax_check_gl.m4, but didn't get accepted.

On the other side, [Chromium BSU, a linux game, not google browser or OS](http://chromium-bsu.sourceforge.net) does use ax_check_gl. but the check failed on OSX 10.9 (just tested). Even if the GL test success the code would still not able to compile because you need to patch configure.ac to make it recognize OpenAL framework on OSX anyway.

peti:
@bastien-roucaries, I am sorry, but the notion of reverting @dryman's efforts makes me feel uncomfortable. If you have concrete and specific criticism about the changes, then I'm sure we can move forward addressing those issues; but just discarding everything he has done in an indiscriminate fashion doesn't feel right.

dryman:
@peti Thanks for the support.

bastien-roucaries:
On Fri, Mar 11, 2016 at 5:15 PM, Felix Chern <notifications@github.com> wrote:
>
> @peti Thanks for the support.


I do not say I do not want the work. I said I prefer the first serie

Bastien

PR #68, opened 2016-03-20
Update CXXCPP after updating CXX.

emcrisostomo:
I'm currently facing this issue: I'm using the `AX_CXX_COMPILE_STDCXX_11` macro (which in turn uses `AX_CXX_COMPILE_STDCXX`) to check for C++11 compiler support.  If C++11 compiler support is detected, this macro sets `CXX` by appending the relevant switch:

```
CXX="$CXX $switch"
```

So far, so good.  Problem is, other variables are evaluated in an earlier stage from `CXX`, such as `CXXCPP` in `AC_PROG_CXXCPP`:

```
[...snip...]
if test -z "$CXXCPP"; then
  AC_CACHE_VAL(ac_cv_prog_CXXCPP,
  [dnl

    # Double quotes because CXXCPP needs to be expanded

    for CXXCPP in "$CXX -E" "/lib/cpp"
    do
      _AC_PROG_PREPROC_WORKS_IFELSE([break])
    done
    ac_cv_prog_CXXCPP=$CXXCPP
  ])dnl

  CXXCPP=$ac_cv_prog_CXXCPP
else
  ac_cv_prog_CXXCPP=$CXXCPP
fi
[...snip...]
```

After a macro sets `CXX` (such as `AX_CXX_COMPILE_STDCXX_11` does), I found no (clean) way to have Autoconf re-evaluate `CXXCPP` again (the check is cached, and anyway calling `AC_PROG_CXXCPP` manually wouldn't be transparent even if it worked).

I think anybody that requests the compiler to be checked and configured to support a certain C++ standard expects the preprocessor to be configured as well.  I you don't, macros such `AC_CHECK_HEADER` wouldn't detect correctly the configuration of the current build and emit warning similar to these one:

```
checking atomic usability... yes
checking atomic presence... no
configure: WARNING: atomic: accepted by the compiler, rejected by the preprocessor!
configure: WARNING: atomic: proceeding with the compiler's result
```

I've patched the aforementioned macros so that they append the same switch to `CXXCPP` after setting `CXX`:

```
CXX="$CXX $switch"
# Warning: /lib/cpp is not being checked (as AC_PROG_CXXCPP does).
if test -n "$CXXCPP" ; then
  CXXCPP="$CXXCPP $switch"
fi
```

I'd rather invoke `AC_PROG_CXXCPP` in order not to duplicate code that manages this kind of variable inter-dependency in multiple places as in:

```
CXX="$CXX $switch"
AC_PROG_CXXCPP
```

but that call it has no effect because the result is cached and not re-evaluated.  Another idea could be cleaning the variables used by AC_PROG_CXXCPP to force re-evaluation but that would be another hack, since you're again coupling a macro with another one and in an even worse way (since you're relying on another macro's internals)

I sent a mail to the Autoconf mailing list but I haven't heard from anybody yet.  Since C++ standards support checks are being integrated into Autoconf, these macros will soon be obsolete and nobody's really caring about them.  However, I thought a patch would be useful.  This has been tested with `g++` and `clang++` on FreeBSD, Debian, Ubuntu, SmartOS, Mac OS X and Cygwin and I found no situations in which setting the preprocessor breaks in any way.

### DIFF
--- a/m4/ax_cxx_compile_stdcxx.m4
+++ b/m4/ax_cxx_compile_stdcxx.m4
@@ -9,9 +9,9 @@
 # DESCRIPTION
 #
 #   Check for baseline language coverage in the compiler for the specified
-#   version of the C++ standard.  If necessary, add switches to CXX to
-#   enable support.  VERSION may be '11' (for the C++11 standard) or '14'
-#   (for the C++14 standard).
+#   version of the C++ standard.  If necessary, add switches to CXX and
+#   CXXCPP to enable support.  VERSION may be '11' (for the C++11 standard)
+#   or '14' (for the C++14 standard).
 #
 #   The second argument, if specified, indicates whether you insist on an
 #   extended mode (e.g. -std=gnu++11) or a strict conformance mode (e.g.
@@ -39,7 +39,7 @@
 #   and this notice are preserved.  This file is offered as-is, without any
 #   warranty.
 
-#serial 3
+#serial 4
 
 dnl  This macro is based on the code from the AX_CXX_COMPILE_STDCXX_11 macro
 dnl  (serial version number 13).
@@ -82,6 +82,9 @@ AC_DEFUN([AX_CXX_COMPILE_STDCXX], [dnl
          CXX="$ac_save_CXX"])
       if eval test x\$$cachevar = xyes; then
         CXX="$CXX $switch"
+        if test -n "$CXXCPP" ; then
+          CXXCPP="$CXXCPP $switch"
+        fi
         ac_success=yes
         break
       fi
@@ -105,6 +108,9 @@ AC_DEFUN([AX_CXX_COMPILE_STDCXX], [dnl
          CXX="$ac_save_CXX"])
       if eval test x\$$cachevar = xyes; then
         CXX="$CXX $switch"
+        if test -n "$CXXCPP" ; then
+          CXXCPP="$CXXCPP $switch"
+        fi
         ac_success=yes
         break
       fi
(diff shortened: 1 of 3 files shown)


emcrisostomo:
Hi @peti, could you explain me why this PR is not building? I've checked the makefiles, especially `cfg.mk` and I think I won't get the exact meaning of this recipe unless I read what those Python scripts are supposed to do:

```
$(STAGEDIR)/%.m4 : $(M4DIR)/%.m4 $(STAGEDIR)/manifest $(srcdir)/macro.py $(srcdir)/macro2m4.py
        $(PYTHON) $(srcdir)/macro2m4.py "$<" "$@"
        @diff -u "$<" "$@" || (touch --date="1970-01-01 01:00:00" "$@"; false)
```

Wouldn't the second line fail just for changing the `#serial`?

peti:
The serial number represents the number of commits that modified the macro. You pushed 2 commits but incremented #serial only once, hence the test suite error. You should probably squash those two commits into one, then everything should be fine.

emcrisostomo:
Hi @peti, thanks.  Apologies for the noise here, but the build failed exactly at the same point when the PR had just 1 commit.  I made the second one and it failed again and it's true the second time I didn't bump the `#serial` (it seemed odd to bump it just for a text rewrap).  Could you have a look at the build log before I make even more noise here?

peti:
The first commit failed, because it didn't word-wrap the documentation properly. Then you fixed that in the second commit, which then didn't bump the #serial number. If you squash those two commits into one, then the test suite should succeed. If in doubt, run "make maintainer - all all" locally to test before pushing a new version.

emcrisostomo:
Hi @peti, thank you very much!  I understood correctly about the text wrapping (hence the second commit), but failed to notice the `#serial` enforcement.  I just squashed the commits and everything's going fine.

emcrisostomo:
Thanks @peti.

peti:
Thank you for the submission!

PR #69, opened 2016-03-20
AX_WITH_CURSES: Enable pkg-config support

SoapGentoo:
Dear Peter,
I've modernized the **AX_WITH_CURSES** macro, which was in dire need of updating. Some of the improvements:

1. We now first try if we can infer the required flags with pkg-config. Information from .pc is more authoritative than just the flags we have been throwing at it. Unfortunately, pkg-config files in ncurses are optional, so we still retain the old fallback detection.
2. The fallback detection was broken with ncurses 6, which has shuffled around symbols and generally requires `-ltinfo` linking for anything more than just `initscr()`. We now also try linking with `-ltinfo` in case linking with `-lncurses{,w}` fails the first time.

This change is rather important, as the whole ecosystem is suffering massive build system breakage due to the required `-ltinfo` flag. I'll try and convince the ncurses guys to enable pkg-config unconditionally, as the optional nature of the .pc files (rightfully) keeps upstream developers from solely relying on pkg-config.

### DIFF
--- a/m4/ax_with_curses.m4
+++ b/m4/ax_with_curses.m4
@@ -12,7 +12,9 @@
 #   present, along with the associated header file.  The NcursesW
 #   (wide-character) library is searched for first, followed by Ncurses,
 #   then the system-default plain Curses.  The first library found is the
-#   one returned.
+#   one returned. Finding libraries will first be attempted by using
+#   pkg-config, and should the pkg-config files not be available, will
+#   fallback to combinations of known flags itself.
 #
 #   The following options are understood: --with-ncursesw, --with-ncurses,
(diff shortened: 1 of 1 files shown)


SoapGentoo:
@peti concerning the travis failure, I can't see the problem, could you help out here? Thanks!

peti:
All macro documentation is supposed to have consistent word-wrapping (at column 72, I believe), and this macro doesn't have that, hence the test suite failure: https://travis-ci.org/peti/autoconf-archive/builds/117338529#L1513. You can run those checks locally as "make maintainer-all all".

SoapGentoo:
@peti I've fixed all issues.

PR #70, opened 2016-03-22
macros for X86 instruction set tests

dryman:
Added two macros: ax_check_x86_features.m4, and ax_gcc_x86_cpu_supports.m4

It checks if the host cpu supports various instruction set: mmx, sse, popcnt,
avx, avx2, etc. If the instruction were supported, it would export
HAVE_XXX_INSTRUCTIONS C preprocessor macros and add the instruction set flag to
X86_FEATURES_CFLAGS variable. By default it would also set CFLAGS, but this can
be disabled by the ACTION-IF-FOUND hook.

The code were tested with real project.


peti:
Thank you very much.

PR #77, opened 2016-05-06
AX_HAVE_QT: Silent make for variable setting

eikel:
Add "-s" to make arguments. Without, make's output (like "Entering directory")
gets assigned to output variables.

### DIFF
--- a/m4/ax_have_qt.m4
+++ b/m4/ax_have_qt.m4
@@ -53,7 +53,7 @@
 #   and this notice are preserved. This file is offered as-is, without any
 #   warranty.
 
-#serial 12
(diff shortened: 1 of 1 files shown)


peti:
Could you please force-push a new  version of this PR that bumps the `#serial` number to fix the test suite failure https://travis-ci.org/peti/autoconf-archive/builds/128243275#L1125? You can also run those tests locally by executing `make maintainer-all all`.

eikel:
I increased the serial number, unfortunately the test still fails.

peti:
Please squash those two commits into one and *force-push* the new verson.

PR #79, opened 2016-05-08
Update ax_with_curses.m4 and ax_with_curses_extra.m4 to match PKG_CHECK_MODULES style

SoapGentoo:
* Ultimately, ax_with_curses.m4 and ax_with_curses_extra.m4 should fall out of
  use once ncurses pkg-config files have become ubiquitous in the ecosystem.
  Until then, the precious variables have been renamed to match the style of
  `PKG_CHECK_MODULES`, such that a later replacement of the macros can be
  performed with a simple sed.

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -10,6 +10,26 @@
 
 * Noteworthy changes in release ?.? (????-??-??) [?]
 
+  AX_WITH_CURSES and AX_WITH_CURSES_* macros have been modernized, in line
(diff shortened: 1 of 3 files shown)


peti:
Since this is an API breaking change, would you mind adding a brief explanation to the `NEWS` file? That way, users of the release archive can quickly find out about the need to adapt their code.

SoapGentoo:
@peti Updated as per your request.

peti:
Awesome, thank you very much.

PR #80, opened 2016-05-27
AX_CHECK_OPENSSL: use AC_CHECK_TOOL for pkg-config

hasufell:
Some distros provide prefixed pkg-config. This patch ensures
the right tool is picked.

See https://www.gnu.org/software/autoconf/manual/autoconf.html#Generic-Programs
for details


peti:
Would you mind force-pushing an amended version of this commit also bumps the `#serial` number?

peti:
@hasufell, ping?

PR #81, opened 2016-06-14
Add AX_CHECK_OPEN62541_{H,LIB} macros

olaf-mandel:
These macros search for the header and library files for the open62541
library [1]. Support both the amalgamation header and the individual
header file layout. Support both the shared and static library, searched
for in a configurable order. Support for adding the library source
directory and the build directory to the search path if the library is
not installed on the system.

[1]: <http://open62541.org/>


peti:
The convention in the Autoconf Archive is that every file publishes exactly one macro and that macro is named like the file. Your code looks very nice and I'd like to merge it, but I wonder whether it would be possible to re-factor the submission to adhere to that convention?

olaf-mandel:
Am 15.06.2016 um 10:21 schrieb Peter Simons:
> The convention in the Autoconf Archive is that every file publishes exactly one macro and that macro is named like the file. [is it] possible to re-factor the submission to adhere to that convention?
> 

The problem is that both macros use the --with-open62541 and
--with-open62541-build options. These are defined in the (up to now)
internal macro _AX_CHECK_OPEN62541_get_paths. To split the two public
macros AX_CHECK_OPEN62541_H and AX_CHECK_OPEN62541_LIB into separate
files, I need to make three files and three public macros:

* AX_CHECK_OPEN62541_H
* AX_CHECK_OPEN62541_LIB
* AX_CHECK_OPEN62541_PATHS

If that is "stylistically better" than keeping the tightly connected
macros in a single file, I can do the refactoring. Please let me know
which alternative you prefer.

peti:
Having these 3 macros split into separate files seems better to me, yes.

peti:
Very cool, thank you!

PR #82, opened 2016-06-15
HDF5 type variable

joelfrederico:
I have a problem where I need to know whether HDF5 was built parallel or serial. I can take care of it in my configure.ac, so I would like the AX_LIB_HDF5 macro to set a variable I can test. The current variables don't reliably give me that information. So I updated the macro to return that information.


joelfrederico:
Could somebody help me understand why my build failed?

peti:
You didn't bump the `#serial` number. You can use "make maintainer-all all" locally to verify your submission before pushing it.

joelfrederico:
@peti Oh just saw this! Thanks! I had some issues with the way I did it too, I've tested some more and now it appears to work

PR #88, opened 2016-08-13
AX_IS_RELEASE: Support out-of-tree builds

tdz:
Signed-off-by: Thomas Zimmermann <tdz@users.sourceforge.net>


peti:
Pushed in ae88d07b44bbd211de37cd5f4cf9877a896b1d52.

PR #90, opened 2016-08-19
AX_CODE_COVERAGE: Rename CODE_COVERAGE_LDFLAGS to CODE_COVERAGE_LIBS

pwithnall:
Since it’s supposed to be added to the Makefile’s target_LIBS variable,
this is a lot clearer.

Continue to support CODE_COVERAGE_LDFLAGS, but it’s deprecated.


peti:
Merged in b812cb0104efaa25ce2f067a950fbf2abc36ebe5.

PR #91, opened 2016-08-29
Fix reference to LIBMYSQLCPPCONN_CXXFLAGS

astoltz:
ax_lib_mysqlcppconn.m4 has a typo referencing CFLAGS. This typo prevents --with-libmysqlcppconn from being used.


peti:
Would you mind force-pushing an amended commit that bumps the `#serial` number of the macro to fix the test suite failure on travis-ci?

PR #93, opened 2016-09-27
AX_CODE_COVERAGE: Define CODE_COVERAGE_LIBS

ptomato:
The previous change seems to have defined LDFLAGS twice rather
than LIBS and LDFLAGS.


ptomato:
@pwithnall Is this correct?

peti:
Looks good to me. :+1:

pwithnall:
Oops. This change looks very good to me.

PR #94, opened 2016-09-27
Add AX_SANITIZERS macro

stapelberg:
PLEASE DO NOT MERGE YET!

Sending this out for early feedback. I have a couple of questions:
- Do you agree in general that this is a useful macro, or are there any concerns about its adoption?
- Please see the TODO in the code about having to specify `-Wc,` for libtool — any pointers/insights as to how to approach that?
- Please see the TODO in the code about specifying `-O1` — does that make sense? Any ideas about how to accomplish that?
- Is there any sort of style guide for formatting m4 code in the context of autoconf archive?

### DIFF
--- a/m4/ax_sanitizers.m4
+++ b/m4/ax_sanitizers.m4
@@ -1,0 +1,97 @@
+# ===========================================================================
+#      http://www.gnu.org/software/autoconf-archive/ax_sanitizers.html
+# ===========================================================================
+#
+# SYNOPSIS
+#
+#   AX_SANITIZERS([SANITIZERS], [ENABLED-BY-DEFAULT], [ACTION-SUCCESS])
+#
+# DESCRIPTION
+#
+#   Offers users to enable one or more sanitizers (see
+#   https://github.com/google/sanitizers) with the corresponding
+#   --enable-<sanitizer>-sanitizer option.
+#
+#   SANITIZERS is a whitespace-separated list of sanitizers to offer via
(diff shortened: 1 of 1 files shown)


peti:
> Do you agree in general that this is a useful macro, or are there any concerns about its adoption?

I am sorry, but I don't feel qualified to judge whether a macros is useful or not. The best I can say is whether a macro is helpful to *me*, but that's not quite the same thing. Generally speaking, the Archive has usually included macros that people had been using before, i.e. that were proven to be useful to someone in practice. What that means for your particular submission is hard to judge. It might be a good idea to contact autoconf-archive-maintainers@gnu.org via e-mail and to ask the "general public" of the archive maintainers for their opinion.

> Please see the TODO in the code about having to specify -Wc, for libtool — any pointers/insights as to how to approach that?

This, too, should probably be addressed to the mailing list. I, personally, don't have time at the moment to look into these matters in sufficient depth to be comfortable giving technical advice on these matters.

> Is there any sort of style guide for formatting m4 code in the context of autoconf archive?

You can format the m4 source code any way you want. There are some checks run by "make maintainer-all all", which are also enforced by the travis-ci test suite enabled for this project, but these checks are only concerned with the documentation and the macro's header, not with the m4 source code itself.

stapelberg:
Thanks for your quick reply.

For the record, I took this discussion to http://lists.gnu.org/archive/html/autoconf-archive-maintainers/2016-10/msg00004.html and will update the pull request once the answers become clear :).

peti:
Please re-submit when ready.

PR #96, opened 2016-10-01
ax_compiler_flags_*: fix underquoting causing infinite recursion

smcv:
These macros would fail with infinite recursion if used twice for the
same variable, for example

    AX_COMPILER_FLAGS_CFLAGS([WARN_CFLAGS], ... some options ...)
    AX_COMPILER_FLAGS_CFLAGS([WARN_CFLAGS], ... more options ...)

In particular, invoking AX_COMPILER_FLAGS, followed by
AX_COMPILER_FLAGS_CFLAGS with more "yes" options has this failure mode.

This is because the first time through the macro, we define
ax_warn_cflags_variable = "WARN_CFLAGS" as expected. The second time,
because the first parameter is underquoted, its value is substituted
before calling m4_define, so we inadvertently define
WARN_CFLAGS = "WARN_CFLAGS". The next time WARN_CFLAGS is mentioned,
attempts to expand it will recurse forever, because m4 does not
special-case a macro that appears in its own expansion like cpp does.

Signed-off-by: Simon McVittie <smcv@debian.org>

---

cc @pwithnall

### DIFF
--- a/m4/ax_compiler_flags_cflags.m4
+++ b/m4/ax_compiler_flags_cflags.m4
@@ -25,7 +25,7 @@
 #   and this notice are preserved.  This file is offered as-is, without any
 #   warranty.
 
-#serial 11
+#serial 12
 
 AC_DEFUN([AX_COMPILER_FLAGS_CFLAGS],[
     AC_REQUIRE([AC_PROG_SED])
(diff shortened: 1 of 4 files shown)


pwithnall:
Whoops. Looks good to me. This was not an anticipated use case of the `AX_COMPILER_FLAGS_*` macros, but it seems legitimate enough.

PR #103, opened 2016-12-24
AX_TYPE_SOCKLEN_T: fix "-Wall -Werror" compatibility

proski:
The len variable is unused, which causes a compile error resulting in a
failure to detect socklen_t.

Also fix a potential warning about implicit type conversion while
initializing len.


proski:
cc macro author @larsbrinkhoff - please review

larsbrinkhoff:
Sorry, I missed the ping.  Looks good to me.

PR #106, opened 2017-01-09
Update ax_have_qt.m4

flightlesson:
Make work on macOS Sierra, where the 'make' commands that set QT_CXXFLAGS and QT_LIBS failed because the embedded qmake wasn't started in the correct directory.


peti:
Can you please force-push a bump of `#serial`  to make the travis-ci tests succeed? You can also run `make maintainer-all all` locally to run the tests on your machine.

peti:
Patch fails regression test suite.

PR #107, opened 2017-01-20
Add new macro AX_ADD_FORTIFY_SOURCE

SoapGentoo:
Hi @peti 
This is a new macro to help the ecosystem not add `-D_FORTIFY_SOURCE=2` unconditionally in their `configure.ac`s. The problem is that some distributions have this flag unconditionally enabled in their toolchains, causing unnecessary churn (or even failure with `-Werror`). On the other hand, building with `-D_FORTIFY_SOURCE=2` on is strongly recommended, in order to catch potential security issues.

### DIFF
--- a/m4/ax_add_fortify_source.m4
+++ b/m4/ax_add_fortify_source.m4
@@ -1,0 +1,53 @@
+# ===========================================================================
+#   http://www.gnu.org/software/autoconf-archive/ax_add_fortify_source.html
+# ===========================================================================
+#
+# SYNOPSIS
+#
+#   AX_ADD_FORTIFY_SOURCE
+#
+# DESCRIPTION
+#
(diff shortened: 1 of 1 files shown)


peti:
Would you consider using the all-permissive license instead of "public domain" for this code? It reads:

> Copying and distribution of this file, with or without modification, are permitted in any medium without royalty provided the copyright notice and this notice are preserved. This file is offered as-is, without any warranty.

This is the choice the FSF recommends to Autoconf macros.

SoapGentoo:
@peti as you wish :wink: Done.

SoapGentoo:
@peti ping?

peti:
Thanks for the submission!

SoapGentoo:
@peti thanks for your commitment and willingness to merge! 😄

PR #110, opened 2017-02-06
switch gnu.org sites to https

vapier:


peti:
Very cool, thank you!

PR #111, opened 2017-02-10
AX_CHECK_ENABLE_DEBUG: Do not conflict with -DNDEBUG

em-:
Since -DNDEBUG is equivalent to "#define NDEBUG 1", explicitly set
the value to ensure that both definitions are equivalent and won't
lead to compiler warnings.

This makes AX_CHECK_ENABLE_DEBUG compatible with AX_CODE_COVERAGE, as
the latter causes -DNDEBUG to be set.

In particular, if AX_COMPILER_FLAGS is also used and thus -Werror
is set (which is the default in that case) this avoids the following
build failure:

./config.h:38:0: error: "NDEBUG" redefined [-Werror]

Submitted in <https://savannah.gnu.org/patch/?9226>

Signed-off-by: Emanuele Aina <emanuele.aina@collabora.com>

### DIFF
--- a/m4/ax_check_enable_debug.m4
+++ b/m4/ax_check_enable_debug.m4
@@ -42,7 +42,7 @@
 #   permitted in any medium without royalty provided the copyright notice
 #   and this notice are preserved.
 
-#serial 6
+#serial 7
 
 AC_DEFUN([AX_CHECK_ENABLE_DEBUG],[
     AC_BEFORE([$0],[AC_PROG_CC])dnl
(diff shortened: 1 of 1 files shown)


peti:
Can you please force-push a new version that bumps the `#serial` number, too? (See the regression test suite error at https://travis-ci.org/peti/autoconf-archive/builds/200374948#L702).

em-:
Force-pushed. :)

em-:
Thanks!

PR #112, opened 2017-02-12
Improve AX_CHECK_ALIGNED_ACCESS_REQUIRED

proski:
Square brackets were overquoted. Avoid square brackets, use quadrigraph
for "#".

Invert the logic so that a crash in the test would count as "yes" for the
aligned address requirement.

When cross-compiling, assume that address alignment is needed, warn the
user.

Use AC_RUN_IFELSE and AC_LANG_PROGRAM, they use sane quoting and take
care of the main function.

Don't include stdio.h, it's not needed for malloc() or anything else.

### DIFF
--- a/m4/ax_check_aligned_access_required.m4
+++ b/m4/ax_check_aligned_access_required.m4
@@ -4,7 +4,7 @@
 #
 # SYNOPSIS
 #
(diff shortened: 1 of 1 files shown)


proski:
@gdraheim: please review.

PR #118, opened 2017-03-22
Bugfixes for AX_RUBY_DEVEL

dividedmind:
- Don't use AX_WITH_RUBY which was removed six years ago.
- Use modern Ruby configuration sources.
- Make the version check actually work.

### DIFF
--- a/m4/ax_ruby_devel.m4
+++ b/m4/ax_ruby_devel.m4
@@ -52,24 +52,24 @@
 #   modified version of the Autoconf Macro, you may extend this special
 #   exception to the GPL to apply to your modified version as well.
 
-#serial 12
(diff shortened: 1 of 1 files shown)


dividedmind:
Thanks for merging. Did you notice the Travis build fail? I'm not sure what it means, but it appears to be related to serial bump. FWIW, I've already bumped the serial in my pull request, so maybe that's the problem. Or maybe you should have bumped it again on merge?

peti:
The serial number corresponds to the  number of commits to the macro's file, so this PR would have needed a double increment -- one per commit. I modified the second commit to bump the serial number, too, before pushing it to the repo.

PR #120, opened 2017-04-10
AX_SUBDIRS_CONFIGURE: new macro

Harenome:
This pull request proposes the new macro `AX_SUBDIRS_CONFIGURE`. This macro provides one with an easy way to customize arguments passed down from a parent `configure` script to its children `configure` scripts.

### `AX_SUBDIRS_CONFIGURE`

`AX_SUBDIRS_CONFIGURE` is a macro which aims to solve the problem of customizing options and arguments for children `configure` scripts while attempting to remain as close as possible to the behaviour of [`AC_CONFIG_SUBDIRS`](https://www.gnu.org/software/autoconf/manual/autoconf-2.63/html_node/Subdirectories.html). It differs from [`AC_CONFIG_SUBDIRS`](https://www.gnu.org/software/autoconf/manual/autoconf-2.63/html_node/Subdirectories.html) in that it accepts 4 additional (and optional) arguments:

- options that must always be passed down to child `configure` scripts.
- options that may be merged with existing options.
- options that must replace existing options.
- options that must never be passed to the `configure` scripts.

Also, directories recorded with `AX_SUBDIRS_CONFIGURE` are added to the variable `subdirs_extra` as opposed to [`AC_CONFIG_SUBDIRS`](https://www.gnu.org/software/autoconf/manual/autoconf-2.63/html_node/Subdirectories.html) which adds its directories to the variable `subdirs`.

Here is a short example where the top project builds projects `A`, `B`, `C` and `D`:
```sh
AC_CONFIG_SUBDIRS([dependencies/A])
AX_SUBDIRS_CONFIGURE(
    [dependencies/B], [--enable-special-feature], [], [],
    [--with-B=build])
AX_SUBDIRS_CONFIGURE(
    [[dependencies/C],[dependencies/D]],
    [],
    [[CPPFLAGS=-I${ac_top_srcdir}/dependencies/B -I${ac_top_builddir}/dependencies/B],
     [LDFLAGS=-L${ac_abs_top_builddir}/dependencies/B/.libs]],
    [--with-B=system],
    [])
```
In the snippet above, the `configure` script for `dependencies/B`, shall always get `--enable-special-feature` and never receive `--with-B=build`. The `configure` scripts for `dependencies/C` and `dependencies/D` will receive `--with-B=system` even if the parent `configure` script had another value for `--with-B`. These scripts will also receive the `CPPFLAGS` and `LDFLAGS` specified as parameters for `AX_SUBDIRS_CONFIGURE`. If the parent `configure` script received values for `CPPFLAGS` or `LDFLAGS`, the children scripts will receive both the values from the parent and the values specified in the macro.

Since one of the goals of the macro is to deviate as less as possible from [`AC_CONFIG_SUBDIRS`](https://www.gnu.org/software/autoconf/manual/autoconf-2.63/html_node/Subdirectories.html), the two following lines should be almost equivalent (the difference is that the first line modifies the variable `subdirs` whereas the second line modifies the variable `subdirs_extra`):
```sh
AC_CONFIG_SUBDIRS([something])
AX_SUBDIRS_CONFIGURE([something])
```

### Related work
How to set specific flags or arguments for children configured with [`AC_CONFIG_SUBDIRS`](https://www.gnu.org/software/autoconf/manual/autoconf-2.63/html_node/Subdirectories.html) is a recurring question which has had several answers over the years:

- directly run the `configure` scripts either in the parent `configure` script or in the `Makefile`. For instance:
    ```sh
    # configure.ac:
    CONFIGURE_ARGS="$ac_configure_args"
    AC_SUBST(CONFIGURE_ARGS)

    # Makefile.am:
    CONFIGURE_ARGS = @CONFIGURE_ARGS@ --some-option
    libsomething:
        cd libsomething && ./configure $(CONFIGURE_ARGS) && make
    ```
- set a variable, `export(1)` it and use [`AC_CONFIG_SUBDIRS`](https://www.gnu.org/software/autoconf/manual/autoconf-2.63/html_node/Subdirectories.html). Doing so may require to “[_un-precious_](http://stackoverflow.com/a/34334802)” the variable of interest.
  An example from StackOverflow **\[[1](http://stackoverflow.com/a/34335084)\]**:
    ```sh
    AC_DEFUN([AX_UNPRECIOUS], [
        m4_define([_AC_PRECIOUS_VARS], m4_bpatsubst(_AC_PRECIOUS_VARS, [$1
    ], []))
    ])
    AX_APPEND_FLAG([-Dtop-configure], [CFLAGS])
    AX_UNPRECIOUS([CFLAGS])
    export CFLAGS
    AC_CONFIG_SUBDIRS([libsomething])
    ```
- modify the value of `$ac_configure_args`. The trick is to use [`AC_CONFIG_COMMANDS_PRE`](https://www.gnu.org/software/autoconf/manual/autoconf-2.67/html_node/Configuration-Commands.html) and [`AC_CONFIG_COMMANDS_POST`](https://www.gnu.org/software/autoconf/manual/autoconf-2.67/html_node/Configuration-Commands.html) to make sure that `config.status` behaves correctly.
  As seen in old versions of Apache Mesos **\[[2](https://github.com/apache/mesos/blob/0.28.2/configure.ac)\]**:
    ```sh
    ac_configure_args_pre="$ac_configure_args"
    ac_configure_args_post="$ac_configure_args --some-option"
    ac_configure_args="$ac_configure_args_post"
    AC_CONFIG_COMMANDS_PRE([ac_configure_args="$ac_configure_args_pre"])
    AC_CONFIG_COMMANDS_POST([ac_configure_args="$ac_configure_args_post"])
    ```
- a patch which introduced the macro `AC_CONFIG_SUBDIR_CUSTOM` was sent to the `autoconf` mailing-lists **\[[3](https://lists.gnu.org/archive/html/autoconf/2011-04/msg00006.html)\]**:
    ```sh
    arguments_for_libsomething="--some-option"
    AC_CONFIG_SUBDIR_CUSTOM([libsomething], [$arguments_for_libsomething])
    ```
- user-contributed macros such as **\[[4](https://neverpanic.de/blog/2014/04/10/autoconf-ac-config-subdirs-with-custom-flags-for-subprojects/)\]**.

The problem with solutions from **\[[1](http://stackoverflow.com/a/34335084)\]** and **\[[2](https://github.com/apache/mesos/blob/0.28.2/configure.ac)\]** is that they require a little bit of knowledge on some of the internals of `autoconf(1)`. Moreover, these solutions make it difficult to use different sets of arguments on multiple `configure` scripts. Manually managing and running children `configure` scripts in `configure` or in a `Makefile` as in the first example may allow to use different sets of options for different `configure` scripts but doing so can quickly lead to bloated glue code. The patch from **\[[3](https://lists.gnu.org/archive/html/autoconf/2011-04/msg00006.html)\]** apparently never made it upstream and thus would require any end-user to use a patched `autoconf(1)` (assuming the patch can still be applied without errors). Lastly the macro from **\[[4](https://neverpanic.de/blog/2014/04/10/autoconf-ac-config-subdirs-with-custom-flags-for-subprojects/)\]** is tailored to specific needs and does more (extracting tarballs) than managing `configure` scripts in subprojects.

**\[[1](http://stackoverflow.com/a/34335084)\]**: http://stackoverflow.com/a/34335084
**\[[2](https://github.com/apache/mesos/blob/0.28.2/configure.ac)\]**: https://github.com/apache/mesos/blob/0.28.2/configure.ac
**\[[3](https://lists.gnu.org/archive/html/autoconf/2011-04/msg00006.html)\]**: https://lists.gnu.org/archive/html/autoconf/2011-04/msg00006.html
**\[[4](https://neverpanic.de/blog/2014/04/10/autoconf-ac-config-subdirs-with-custom-flags-for-subprojects/)\]**: https://neverpanic.de/blog/2014/04/10/autoconf-ac-config-subdirs-with-custom-flags-for-subprojects/

### DIFF
--- a/m4/ax_subdirs_configure.m4
+++ b/m4/ax_subdirs_configure.m4
@@ -1,0 +1,325 @@
+# NAME
+#
+#   AX_SUBDIRS_CONFIGURE - AC_CONFIG_SUBDIRS with customizable options
+#
+# SYNOPSIS
+#
+#   AX_SUBDIRS_CONFIGURE(
+#       [subdirs],
+#       [mandatory arguments],
+#       [possibly merged arguments],
+#       [replacement arguments],
+#       [forbidden arguments])
+#
+# SUMMARY
+#
+#   AX_SUBDIRS_CONFIGURE attempts to be the equivalent of AC_CONFIG_SUBDIRS
+#   with customizable options for configure scripts.
+#
+# DESCRIPTION
+#
+#   Run the configure script for each directory from the comma-separated m4 list
+#   'subdirs'. This macro can be used multiple times. All arguments of this
+#   macro must be comma-separated lists.
+#
+#   All command line arguments from the parent configure script will be given
+#   to the subdirectory configure script after the following modifications (in
+#   that order):
+#
+#   1. The arguments from the 'mandatory arguments' list shall always be
+#      appended to the argument list.
+#   2. The arguments from the 'possibly merged arguments' list shall be added
+#      if not present in the arguments of the parent configure script or
+#      merged with the existing argument otherwise.
+#   3. The arguments from the 'replacement arguments' list shall be added
+#      if not present in the arguments of the parent configure script or
+#      replace the existing argument otherwise.
+#   4. The arguments from the 'forbidden arguments' list shall always be removed
+#      from the argument list.
+#
+#   The lists 'mandatory arguments' and 'forbidden arguments' can hold any kind
+#   of argument. The 'possibly merged arguments' and 'replacement arguments'
+#   expect their arguments to be of the form --option-name=value.
+#
+#   This macro aims to remain as close as possible to the AC_CONFIG_SUBDIRS
+#   macro. It corrects the paths for '--cache-file' and '--srcdir' and adds
+#   '--disable-option-checking' and '--silent' if necessary.
+#
+#   This macro also sets the output variable subdirs_extra to the list of
+#   directories recorded with AX_SUBDIRS_CONFIGURE. This variable can be
+#   used in Makefile rules or substituted in configured files.
+#
+# NOTES
+#
+#   This macro shall do nothing more than managing the arguments of the
+#   configure script. Just like when using AC_CONFIG_SUBDIRS, it is up to
+#   the user to check any requirements or define and substitute any required
+#   variable for the remainder of the project.
+#
+#   Configure scripts recorded with AX_SUBDIRS_CONFIGURE may be executed before
+#   configure scripts recorded with AC_CONFIG_SUBDIRS.
+#
+#   Without additional arguments, the behaviour of AX_SUBDIRS_CONFIGURE should
+#   be identical to the behaviour of AC_CONFIG_SUBDIRS, apart from the contents
+#   of the variables subdirs and subdirs_extra (except that AX_SUBDIRS_CONFIGURE
+#   expects a comma-separated m4 list):
+#
+#       AC_CONFIG_SUBDIRS([something])
+#       AX_SUBDIRS_CONFIGURE([something])
+#
+#   This macro may be called multiple times.
+#
+# EXAMPLE
+#
+#   Let us assume our project has 4 dependencies, namely A, B, C and D. Here
+#   are some characteristics of our project and its dependencies:
+#   - A does not require any special option.
+#   - we want to build B with an optional feature which can be enabled with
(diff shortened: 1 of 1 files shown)


peti:
I merged this macro in 2192c6c102c7e3c4e5cd69587ce7bf44ae9f728e and also did some minor formatting changes in 961dd75066bf2065c7cb2af02f6d4e835f067fdb to ensure it conforms to the Archive's guidelines. Thank you very much for this submission!

PR #123, opened 2017-06-15
AX_RUBY_EXT: Use RbConfig instead of deprecated Config module

marv:
Ruby 2.1 (which Debian Jessie ships) warns about the deprecation:

```
# ruby -rrbconfig -e 'puts Config::CONFIG["sitedir"]'
-e:1:in `<main>': Use RbConfig instead of obsolete and deprecated Config.
-e:1:in `<main>': Use RbConfig instead of obsolete and deprecated Config.
/usr/local/lib/site_ruby
```

And newer Ruby versions throw an error:
```
# ruby2.2 -rrbconfig -e 'puts Config::CONFIG["sitedir"]'                               
-e:1:in `<main>': uninitialized constant Config (NameError)
```

### DIFF
--- a/m4/ax_ruby_ext.m4
+++ b/m4/ax_ruby_ext.m4
@@ -56,7 +56,7 @@
 #   ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF
 #   THE POSSIBILITY OF SUCH DAMAGE.
 
-#serial 3
+#serial 4
 
 AC_DEFUN([AX_RUBY_EXT],[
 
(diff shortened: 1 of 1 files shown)


peti:
Will `RbConfig` still work with older versions of Ruby?

marv:
I wasn't really sure which Ruby versions are supposed to be supported and since
AX_RUBY_DEVEL uses RbConfig unconditionally I thought it'd be okay to do the
same in AX_RUBY_EXT.
I just check Ruby 1.8 & 1.9 and they both work with RbConfig:

```
# ruby1.8 -rrbconfig -e "puts RbConfig::CONFIG['sitedir']"
/usr/local/lib/site_ruby
```
```
# ruby1.9.1 -rrbconfig -e "puts RbConfig::CONFIG['sitedir']"
/usr/local/lib/site_ruby
```

peti:
OK, cool. Thank you very much for your efforts!

PR #130, opened 2017-08-28
Fix AX_LIB_ORACLE_OCI macro for plain C

maxsatula:
1. Use C compiler, not C++.
2. Split LDFLAGS into LDFLAGS and LIBS.
   However, merge them back to ORACLE_OCI_LDFLAGS afterwards
   to preserve interface backward compatibility.
3. Replace C++ style comments with C style.

Resolves mloskot/autotools-modules#2


maxsatula:
Cancelled in favor to #132

PR #143, opened 2018-01-10
AX_PKG_CHECK_MODULES: Normalize whitespace in package lists

ptomato:
It can make configure.ac files more readable to list the packages on
separate lines. Strip the newlines and extra whitespace with
m4_normalize.


ptomato:
cc @pwithnall

pwithnall:
Looks OK to me.

PR #150, opened 2018-03-18
fix header file path and version check conditional

twdroeger:
This pull request is a proposed fix for m4/ax_check_mysql.m4.

The changes fixes two problems I encountered when running the configure script.

The first problem is configure is incorrectly checking for the header path using /usr/include/mysql/mysql_version.h/mysql_version.h as the path.

$ ./configure
...
checking /usr/include/mysql/mysql_version.h/mysql_version.h usability... no
checking /usr/include/mysql/mysql_version.h/mysql_version.h presence... no
checking for /usr/include/mysql/mysql_version.h/mysql_version.h... no
....

After fixing the previous problem, the configure script now encounters an invalid command.

checking if MySQL install supports Plugins... yes
checking if MySQL version is equal or greater than 5.7.21... ./configure: line 4958: 21: command not found
yes

### DIFF
--- a/m4/ax_check_mysql.m4
+++ b/m4/ax_check_mysql.m4
@@ -63,7 +63,7 @@
 #   and this notice are preserved. This file is offered as-is, without any
 #   warranty.
 
-#serial 6
+#serial 8
 
 AC_ARG_WITH(mysql,AS_HELP_STRING([--with-mysql],[root of the MySQL installation]))
 AC_ARG_WITH(mysql_plugin,AS_HELP_STRING([--with-mysql-plugin],[path to the MySQL installation plugin directory]))
--- a/m4/ax_check_mysql.m4
+++ b/m4/ax_check_mysql.m4
@@ -118,9 +118,9 @@ AC_DEFUN([AX_CHECK_MYSQL_INSTALL],[
     ROOT_DIR="$1"
 
     # Check for include directory
-    AX_CHECK_MYSQL_INCLUDES([$ROOT_DIR/include/mysql/mysql_version.h])
+    AX_CHECK_MYSQL_INCLUDES([$ROOT_DIR/include/mysql])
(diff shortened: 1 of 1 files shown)


peti:
Pinging @toumorokoshi @twdroeger @vapier @jsoref. Do you have any opinions on this modification?

vapier:
both the `test` and `AX_CHECK_MYSQL_INCLUDES` changes lgtm

the serial change should be squashed in, but i guess you can do that when merging.  not sure why it jumps by 2 when going from 6 to 7 would be fine, but it's also not a big deal as long as it's newly unique.

jsoref:
My guess is that he was following the serial number change style from twdroeger/autoconf-archive@69bdbd1 which was cleanup for my work. It's likely that there were X jumps per file based on the number of times my commits changed each file.

Given that my work involved arbitrary changes (which could be reordered / dropped), it was a bad idea for me to include the serial number changes individually (lots of pain, minimal benefit). But, my changes aren't the typical case (here), where it's a single change to a single script that should get a single `+1` version bump.

twdroeger:
The serial number needed to be bumped up by 2 because I made two commits. The first commit contained my changes with no change to the serial number and the second commit was a change to the serial number.  In the Pull Request, Github displays my changes in the two commits together so that it appears as a single change which may mislead you into thinking that the serial number needs to be bumped by 1.

Initially, I was not aware that the serial number needed to be changed. I only realized this when I noticed that my changes had failed the "Travis-CI build". After analyzing the build failure and the build scripts, I realized that I needed to update the serial number. To confirm my conclusion, I did look at other commits and did notice that at times it was bump by 1 and other times by 2 or more.

jsoref:
fwiw, instead of creating a second commit, you would have wanted to use an `amend` or `squash` operation...

PR #152, opened 2018-04-19
ax_cxx_compile_stdcxx.m4: Drop the check for default C++ version support

krnowak:
This is the same thing as in https://savannah.gnu.org/patch/index.php?9572, but I guess patches on github are processed faster. Please let me know which way is preferred.

The problem with the check is that it can't detect whether the version
is with extensions or not. So it can't take the "noext" or "ext"
option into account.

This may be a problem when an extended version of C++ has a feature
that conflicts with another feature from the C++ standard and we have
no way to tell the macro to add a "-std" flag anyway.

An example:

    using namespace std::complex_literals;
    std::complex<double> c = 1.0;
    c += 1.0i;
    std::cout << "abs" << c << " = " << abs(c) << '\n';

Here, the "1.0i" constant is either a GCC extension if compiled with
-std=gnu++14 or a standard C++ operator if compiled with
-std=c++14. If a GCC extension is used, the snippet fails to
compile. Unfortunately, recent GCC versions default to gnu++14, so
this code fails to compile by default.

Dropping the check forces the macro to evaluate compiler flags with
taking the "ext"/"noext" option into account.

Issue spotted by Frédéric Mangano-Tarumi.

### DIFF
--- a/m4/ax_cxx_compile_stdcxx.m4
+++ b/m4/ax_cxx_compile_stdcxx.m4
@@ -33,14 +33,14 @@
 #   Copyright (c) 2014, 2015 Google Inc.; contributed by Alexey Sokolov <sokolov@google.com>
 #   Copyright (c) 2015 Paul Norman <penorman@mac.com>
 #   Copyright (c) 2015 Moritz Klammler <moritz@klammler.eu>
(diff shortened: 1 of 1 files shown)


peti:
Github is easier for me, indeed. Thank you very much for the submission!

PR #158, opened 2018-05-21
ax_lib_postgresql: Double-quote the literal with square brackets 

alexanderlaw:


m4/ax_lib_postgresql.m4
```diff
@@ -50,7 +50,7 @@
 #   and this notice are preserved. This file is offered as-is, without any
 #   warranty.
 
-#serial 21
+#serial 22
```
peti: This is not going to work. The serial number is the number of commits that contributed to this file. So when you make *two* commits to this file -- like this PR does -- then it must be bumped twice, too. The somewhat nicer alternative is to make both changes in *one commit*. See the "amend" or "squash" features in Git, which allows you to do this kind of thing nicely.

Anyhow, I can squash-merge this PR, which should be fine. Thank you for your contribution!

alexanderlaw: I didn't use amend because I wanted to have a minimal commit that does exactly what it declares. Thanks for clarifications!

PR #147, opened 2018-01-15
ax_compiler_flags_cflags: c++ fixes

rurban:
Several warning flags are incompatible with c++:
-Wnested-externs -Wmissing-prototypes -Wstrict-prototypes -Wdeclaration-after-statement
-Wold-style-definition -Wimplicit-function-declaration -Wjump-misses-init

PR #159, opened 2018-06-07
ax_compiler_flags_cflags: fix disabling C-only warnings

wjt:
https://github.com/autoconf-archive/autoconf-archive/pull/147 introduced
a regression where, because -Wdeclaration-after-statement etc. are
appended to the compile flags after the user-supplied list of warnings
to disable, it is impossible to disable these warnings.


wjt:
Done.

PR #160, opened 2018-06-30
AX_CHECK_COMPILE_FLAG: released under all-permissive license

stweil:
Replace GPL3 + Autoconf exception by all-permissive license.

See discussion at https://github.com/tesseract-ocr/tesseract/pull/1559.

Signed-off-by: Stefan Weil <sw@weilnetz.de>

PR #164, opened 2018-07-22
all-permissive for macros from Guido Draheim / Maarten Bosmans

gdraheim:
There are 5 files with an authorship to @mkbosmans in the autoconf-archive that come from a single commit in 2011. 

git show --summary cfea4ee52f7f2490937976b828a3b82d07b431f2

Author: Maarten Bosmans <mkbosmans@gmail.com>
Date:   Sat Jul 2 12:06:31 2011 +0200

    ax_append_flag: Split file into one macro per file

 create mode 100644 m4/ax_append_compile_flags.m4
 create mode 100644 m4/ax_append_flag.m4
 create mode 100644 m4/ax_append_link_flags.m4
 create mode 100644 m4/ax_check_compile_flag.m4
 delete mode 100644 m4/ax_check_flag.m4
 create mode 100644 m4/ax_check_link_flag.m4
 create mode 100644 m4/ax_check_preproc_flag.m4

I would like to relicense my own stuff to all-permissive. We had some questions about licensing issues from other projects lately, and an easier license makes it easier for them to ship a copy. Compare #160 

I guess that Maarten is okay with that.

### DIFF
--- a/m4/ax_append_compile_flags.m4
+++ b/m4/ax_append_compile_flags.m4
@@ -30,33 +30,12 @@
 #
 #   Copyright (c) 2011 Maarten Bosmans <mkbosmans@gmail.com>
 #
-#   This program is free software: you can redistribute it and/or modify it
-#   under the terms of the GNU General Public License as published by the
-#   Free Software Foundation, either version 3 of the License, or (at your
-#   option) any later version.
-#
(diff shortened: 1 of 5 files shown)


peti:
I reckon we should give @mkbosmans a few days to express his opinion, but if in doubt I tend to assume that everything will be fine, too. :-)

PR #170, opened 2018-07-26
Replace AC_TRY_COMPILE and AC_LANG_

petk:
Autoconf 2.50 in 2001 made several macros obsolete. These include macros for temporary changing the language - `AC_LANG_SAVE`, `AC_LANG_CPLUSPLUS`, `AC_LANG_C`, and `AC_LANG_RESTORE`. Instead of these the `AC_LANG_PUSH` and `AC_LANG_POP` macros should be used with later Autoconf versions.

`AC_TRY_COMPILE` macros should be replaced with AC_COMPILE_IFELSE.

Refs:
- http://git.savannah.gnu.org/cgit/autoconf.git/tree/NEWS
- https://www.gnu.org/software/autoconf/manual/autoconf-2.69/html_node/Language-Choice.html
- https://www.gnu.org/software/autoconf/manual/autoconf-2.69/html_node/Obsolete-Macros.html

### DIFF
--- a/m4/ax_cflags_force_c89.m4
+++ b/m4/ax_cflags_force_c89.m4
@@ -34,15 +34,14 @@
 #   and this notice are preserved.  This file is offered as-is, without any
 #   warranty.
 
-#serial 11
+#serial 12
 
 AC_DEFUN([AX_CFLAGS_FORCE_C89],[dnl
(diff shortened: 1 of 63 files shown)


petk:
Rechecked and merged...

PR #171, opened 2018-07-27
Replace obsolete AC_TRY_COMPILE with AC_COMPILE_IFELSE

petk:
Autoconf made several macros obsolete including the `AC_TRY_COMPILE` in 2000 and since Autoconf 2.50:
http://git.savannah.gnu.org/cgit/autoconf.git/tree/ChangeLog.2

These macros should be replaced with the current `AC_COMPILE_IFELSE` instead.

This patch was created with the help of autoupdate script.

Refs:
- https://www.gnu.org/software/autoconf/manual/autoconf-2.69/html_node/Obsolete-Macros.html
- https://www.gnu.org/software/autoconf/manual/autoconf-2.69/html_node/AC_005fACT_005fIFELSE-vs-AC_005fTRY_005fACT.html

### DIFF
--- a/m4/ax_c_declare_block.m4
+++ b/m4/ax_c_declare_block.m4
@@ -61,13 +61,13 @@
 #   modified version of the Autoconf Macro, you may extend this special
 #   exception to the GPL to apply to your modified version as well.
 
-#serial 10
+#serial 11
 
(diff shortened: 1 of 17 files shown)


petk:
I'm merging this one after checking changes out. (serials were all bumped), and macro syntax should be ok also.

PR #175, opened 2018-08-23
ax_gcc_archflag: include POWER7 -> POWER10

grooverdan:
POWER6 is missing - couldn't find hardware to test (and it was
a long time ago).

Tested on P7 - P9.

Added POWER10 as its in development and unlikely to change


grooverdan:
Thanks @peti

PR #177, opened 2018-11-14
ax_check_enable_debug: Correct NDEBUG regression.

ossama-othman:
Reintroduce change (6e6c447) that addressed NDEBUG redefinition when
using both the AX_CODE_COVERAGE and AX_CHECK_ENABLE_DEBUG macros.  It
was lost in the previous patch to ax_check_enable_debug.m4 (d049eeb).


petk:
Thank you @ossama-othman

ossama-othman:
@petk It was my pleasure.  The Autoconf Archive macros have been very useful to me over the years so it was about time that I contributed. :)

PR #178, opened 2018-12-29
AX_ASM_INLINE: accept asm also

rurban:


m4/ax_asm_inline.m4
```diff
@@ -23,13 +23,13 @@
...
 
-#serial 4
+#serial 5
 
 AC_DEFUN([AX_ASM_INLINE], [
   AC_LANG_PUSH([C])
   AC_MSG_CHECKING(for inline assembly style)
   AC_CACHE_VAL(ac_cv_asm_inline, [
-    ax_asm_inline_keywords="__asm__ __asm none"
+    ax_asm_inline_keywords="__asm__ __asm asm none"
```
peti: Why is this necessary? Would you mind what use case this change solves?

rurban:
`asm` is in Annex J of the C Standard: `J.5.10 The asm keyword`
I haven't checked all esoteric compilers, but most code now uses just `asm`, not `__asm__`

PR #187, opened 2019-04-21
ax_tls.m4: bug fix and language update 

jannick0:


m4/ax_tls.m4
```diff
@@ -44,31 +44,28 @@
...
+    done ]
+  )
   AC_MSG_RESULT([$ac_cv_tls])
 
   AS_IF([test "$ac_cv_tls" != "none"],
-    [AC_DEFINE_UNQUOTED([TLS],[$ac_cv_tls],[If the compiler supports a TLS storage class define it to that here])
-     m4_ifnblank([$1],[$1])],
-    [m4_ifnblank([$2],[$2])])
+    [AC_DEFINE_UNQUOTED([TLS],[$ac_cv_tls],[If the compiler supports a TLS storage class, define it to that here])
+     m4_ifnblank([$1],[$1],[[:]])],
```
petk: Hello, thank you for the patch. This additional is required? Because, I think the `m4_ifnblank` is expanded when creating configure script and not in the configure's shell code directly afterwards.

petk: Ah, sorry, this is inside the AS_IF call... Of course.

petk:
Applied via 474c8f1a4389b7b9dd1ca9b040f7ff0cd7c80132 Thank you @jannick0  for the patch 👍

jannick0:
Great, many thanks.

I am just wondering if this case is just a very special one which should be properly handled by `AS_IF` (defined in [m4sh.ms](http://git.savannah.gnu.org/cgit/autoconf.git/tree/lib/m4sugar/m4sh.m4#n618)).  I can see some amending commits for this over the last couple of years which might cover it, but the last official `autoconf` release was in 2012, a long time ago.

PR #190, opened 2019-05-22
ax_have_qt: workaround qmake bug when building out of tree

rixed:
Makefile generated by qmake include paths relative to the build
directory that is assumed to be current.

### DIFF
--- a/m4/ax_have_qt.m4
+++ b/m4/ax_have_qt.m4
@@ -54,7 +54,7 @@
 #   and this notice are preserved. This file is offered as-is, without any
 #   warranty.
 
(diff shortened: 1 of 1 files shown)


petk:
Thank you for the pull request and the fix. Can you please also update only the serial number (simple internal versioning of these macros):

#serial 16 should be now #serial 17

Other than that I think all good here.

rixed:
Done.

Have a nice day!

rixed:
@peti, @petk Yes apparently I force-pushed the same commit :-/
Should be fixed.

PR #192, opened 2019-07-13
ax_boost_base: fix libsubdirs preference for MIPS64

xen0n:
Currently there're 3 well-known ABIs in use on MIPS64 systems, with
the respective libdir as follows:

* n64: `lib64`,
* n32: `lib32`,
* o32: `lib`.

Probe all three of them.


petk:
LGTM. Tests passed, fix looks good. Thank you @xen0n for helping out

PR #194, opened 2019-07-17
ax_{blas,lapack}: add macOS dynamic library suffix (`.dylib`) globs to `with_{blas,lapack}` case statements  

goxberry:
The `ax_blas` and `ax_lapack` macros use `case $with_blas` and `case $with_lapack` statements to check for a number of library suffixes via globs. Prior to this PR, these suffixes did not include the globs `*.dylib` and `*.dylib.*`, both of which check for macOS suffixes used to denote shared libraries. This PR adds globs corresponding to these macOS shared library suffixes.

### DIFF
--- a/m4/ax_blas.m4
+++ b/m4/ax_blas.m4
@@ -36,6 +36,7 @@
 # LICENSE
 #
 #   Copyright (c) 2008 Steven G. Johnson <stevenj@alum.mit.edu>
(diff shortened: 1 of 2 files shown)


goxberry:
Updated serial versions for `ax_blas.m4` and `ax_lapack.m4`.

PR #195, opened 2019-08-30
Improve message for openssl lookup

vhquang:
This PR originates from https://github.com/python/cpython/pull/15556, where it was suggested to bring it up here.
I was in a situation where I have many install of OpenSSL in many environments, and not all of my install of OpenSSL will be compatible for what I try to do. If we can improve the message with narrower scope of what it is looking, it would help narrowing down my hunt.

Please let me know if anything can be improved on this PR!

### DIFF
--- a/m4/ax_check_openssl.m4
+++ b/m4/ax_check_openssl.m4
@@ -32,7 +32,7 @@
 #   and this notice are preserved. This file is offered as-is, without any
 #   warranty.
 
-#serial 10
+#serial 11
 
 AU_ALIAS([CHECK_SSL], [AX_CHECK_OPENSSL])
 AC_DEFUN([AX_CHECK_OPENSSL], [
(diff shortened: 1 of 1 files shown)


petk:
Hello, thank you for the quick fix. I think that's ok, yes. Can you maybe also bump the serial number of this file (it serves as a quick versioning of changes):

```patch
-#serial 10
+#serial 11
```

vhquang:
petk, thanks for guiding me. I bump the version as your suggestion. Let me know if anything else I would need to do.

peti:
@vhquang, the serial number is the number of commits that modified the macro. This means that the second commit you've added needed yet another bump the serial number and now the test suite fails just the same as before. Can you please squash those two commits into one?

vhquang:
@peti thank you for patiently guiding me and explain to me why the CI test fail. That makes sense and helps a lot! Please let me know if anything I can improve on the PR. I have learnt couple interesting things while doing this PR.

PR #198, opened 2019-10-02
AX_BOOST_BASE: fix Boost detection on some platforms

fanquake:
When the libpath doesn't line up with the value from config.sub, we don't find
the correct path to boost's libs. For example, uname may report `armv7l` while 
libs are installed to `/usr/lib/arm-linux-gnueabihf`. This trys another path 
before giving up.

We've been using [this patch](https://github.com/bitcoin/bitcoin/commit/54c7df81f3e5f81cb91646acaf82074a3a6be3b2) in our repo for some time, but accidentally lost it when we [last updated the `boost_base.m4`](https://github.com/bitcoin/bitcoin/pull/15461), resulting in Boost being reported as [unavailable on some systems](https://github.com/bitcoin/bitcoin/issues/17010).

### DIFF
--- a/m4/ax_boost_base.m4
+++ b/m4/ax_boost_base.m4
@@ -33,7 +33,7 @@
 #   and this notice are preserved. This file is offered as-is, without any
 #   warranty.
 
-#serial 48
+#serial 49
 
 # example boost program (need to pass version)
 m4_define([_AX_BOOST_BASE_PROGRAM],
@@ -133,7 +133,7 @@ AC_DEFUN([_AX_BOOST_BASE_RUNDETECT],[
     libsubdirs="lib/`$CXX -dumpmachine 2>/dev/null` $libsubdirs"
 
     dnl first we check the system location for boost libraries
-    dnl this location ist chosen if boost libraries are installed with the --layout=system option
+    dnl this location is chosen if boost libraries are installed with the --layout=system option
     dnl or if you install boost with RPM
     AS_IF([test "x$_AX_BOOST_BASE_boost_path" != "x"],[
         AC_MSG_CHECKING([for boostlib >= $1 ($WANT_BOOST_VERSION) includes in "$_AX_BOOST_BASE_boost_path/include"])
--- a/m4/ax_boost_base.m4
+++ b/m4/ax_boost_base.m4
@@ -33,7 +33,7 @@
 #   and this notice are preserved. This file is offered as-is, without any
 #   warranty.
 
-#serial 47
+#serial 48
 
 # example boost program (need to pass version)
 m4_define([_AX_BOOST_BASE_PROGRAM],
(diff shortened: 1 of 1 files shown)


m4/ax_boost_base.m4
```diff
@@ -126,8 +126,14 @@ AC_DEFUN([_AX_BOOST_BASE_RUNDETECT],[
       [multiarch_libsubdir="lib/${host_cpu}-${host_os}"]
     )
 
+    dnl some arches may advertise a cpu type that doesn't line up with their
+    dnl prefix's cpu type. For example, uname may report armv7l while libs are
+    dnl installed to /usr/lib/arm-linux-gnueabihf. Try getting the compiler's
+    dnl value for an extra chance of finding the correct path.
+    libsubdirs="lib/`$CXX -dumpmachine 2>/dev/null` $libsubdirs"
```
peti: I doubt that this portable. GNU GCC will respond properly to that command line option, but what about other compilers?

m4/ax_boost_base.m4
```diff
@@ -33,7 +33,7 @@
 #   and this notice are preserved. This file is offered as-is, without any
 #   warranty.
 
-#serial 47
+#serial 49
```
laanwj: Why skip 48?

peti: The serial number of each macros is the number of commits that modified it in the repository.

laanwj: Oh, thanks!

m4/ax_boost_base.m4
laanwj: FWIW clang 10 does as well, at least
```
$ /opt/clang10/bin/clang -dumpmachine
x86_64-unknown-linux-gnu
```
but I agree that this needs error handling, the output of `-dumpmachine` should only be added when the command succeeds

greenaddress: ndk (clang based android native dev kit) doesn't seem to work on 0.19.0.1 (but does on 0.18.1) - testing a bit

peti:
I'm not sure how to proceed with this PR. The current state does not appear to be portable, so I'd rather not apply the patch in this form (even though it clearly fixes a real problem for some users).

luke-jr:
Why does this need to get the library path in the first place...? All this complexity is ultimately just looking for them in the presumed-default system paths anyway. Shouldn't it be able to just try `-lboost_*`?

m4/ax_boost_base.m4
hebasto: This is not portable on ARMv7 with clang:
```
$ clang++ --version
clang version 6.0.0-1ubuntu2 (tags/RELEASE_600/final)
Target: armv8l-unknown-linux-gnueabihf
Thread model: posix
InstalledDir: /usr/bin
$ clang++ -dumpmachine
armv8l-unknown-linux-gnueabihf
```

For comparison:
```
$ g++ --version | head -1
g++ (Ubuntu/Linaro 7.5.0-3ubuntu1~18.04) 7.5.0
$ g++ -dumpmachine
arm-linux-gnueabihf
```

```

peti:
Closing, because of inactivity.

luke-jr:
Your own inactivity? 🤨

luke-jr:
(Found #214 apparently fixing this)

PR #200, opened 2019-10-19
fix AX_ADD_FORTIFY_SOURCE for msys2

rurban:
msys2 just dropped _FORTIFY_SOURCE support.
See https://github.com/msys2/MINGW-packages/issues/5803
Try to actually link it.


rurban:
sorry, the new probe does not catch the new mingw64 problem yet. I don't have any windows machine, only appveyor. so this could need more time. will resubmit later

PR #203, opened 2019-11-11
Use want_boost_process instead of want_boost

Sjors:
I missed this in #202: `want_boost` is ambiguous, so this PR sets `want_boost_process` based on `--with-boost-process`

### DIFF
--- a/m4/ax_boost_process.m4
+++ b/m4/ax_boost_process.m4
@@ -31,7 +31,7 @@
 #   and this notice are preserved. This file is offered as-is, without any
 #   warranty.
 
-#serial 1
+#serial 2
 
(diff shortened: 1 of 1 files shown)


petk:
Sounds good. Thank you @Sjors

Sjors:
Thanks for merging @petk. Is there anything that needs to happen for https://www.gnu.org/software/autoconf-archive/ax_boost_process.html to work? It's still a 404.

petk:
Probably the upstream Git repository http://git.savannah.gnu.org/cgit/autoconf-archive.git/ should be synced with GitHub repo... It's done manually based on activity I think.

PR #205, opened 2019-11-28
ax_have_epoll: avoid -Wunused-but-set-variable errors

simark:
When building with -Werror -Wunused-but-set-variable, the code snippets
in this file generate a warning, causing the test to spuriously fail.
Remove the "rc" variables, since they're not used.

Sample config.log output:

    configure:22489: checking for Linux epoll(7) interface
    configure:22515: ccache gcc -o conftest -g3 -O0 -Wall -Werror -Wextra -Wno-type-limits -Wno-missing-declarations -Wno-missing-field-initializers -Wno-si      gn-compare -Wno-implicit-fallthrough -Wno-unused-parameter -I/home/smarchi/install/include -DHAVE_LINUX_VERSION_H -L/home/smarchi/install/lib conftest.c       -lurcu-cds  >&5
    conftest.c: In function 'main':
    conftest.c:184:9: error: variable 'rc' set but not used [-Werror=unused-but-set-variable]
     int fd, rc;
             ^~
    cc1: all warnings being treated as errors

### DIFF
--- a/m4/ax_have_epoll.m4
+++ b/m4/ax_have_epoll.m4
@@ -42,7 +42,7 @@
 #   and this notice are preserved. This file is offered as-is, without any
 #   warranty.
 
-#serial 11
+#serial 12
 
 AC_DEFUN([AX_HAVE_EPOLL], [dnl
   ax_have_epoll_cppflags="${CPPFLAGS}"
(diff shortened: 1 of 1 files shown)


simark:
Ping.

petk:
Thank you @simark :+1:

PR #206, opened 2019-12-02
Don't check the __cplusplus version on xlC

cbiesinger:
It always defines it as 199711L. See:
https://www.ibm.com/support/knowledgecenter/en/SSLTBW_2.3.0/com.ibm.zos.v2r3.cbclx01/cplr367.htm
(contrast with the documentation for __STDC_VERSION__)

Test output with xlC 16.1.0:
bash-4.4$ /opt/IBM/xlC/16.1.0/bin/xlC test.cc
bash-4.4$ ./a.out
version 199711
bash-4.4$ /opt/IBM/xlC/16.1.0/bin/xlC test.cc -qlanglvl=extended0x
bash-4.4$ ./a.out
version 199711
bash-4.4$ cat test.cc
int main() {
  fprintf(stderr, "version %li\n", __cplusplus);
}

### DIFF
--- a/m4/ax_cxx_compile_stdcxx.m4
+++ b/m4/ax_cxx_compile_stdcxx.m4
@@ -165,7 +165,8 @@ m4_define([_AX_CXX_COMPILE_STDCXX_testbody_new_in_11], [[
 
 #error "This is not a C++ compiler"
 
-#elif __cplusplus < 201103L
+// xlC always defines __cplusplus as 199711L, even when using a later version.
+#elif __cplusplus < 201103L && !defined(__xlC__)
 
 #error "This is not a C++11 compiler"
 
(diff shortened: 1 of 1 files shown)


cbiesinger:
Although, it looks like even the latest version of the compiler does not support enough C++11 for this to work out (it also requires a different flag). And it looks like they may be switching to a clang-based frontend which does not have the issue. So not sure if this patch is worth it

peti:
Are you still interested in this patch? I wonder because it fails the test suite -- probably because `#serial` wasn't bumped.

cbiesinger:
Yeah, I'll abandon this, the compiler is missing a number of C++11 features that make it fail the rest of the checks.

PR #209, opened 2019-12-25
Add ax_restrict

rurban:
to check for broken gcc-5 and clang-6 implementations, and undefine restrict then.


peti:
Closed because of inactivity.

PR #211, opened 2019-12-31
ax_boost_thread: do not set -pthread on Android

icota:
The android libc, bionic, provides built-in support for pthreads, so no additional linking is necessary.
Setting `-pthread` [breaks the build](https://github.com/protocolbuffers/protobuf/issues/1373).


petk:
Thank you @icota for the Android patch.

PR #212, opened 2020-01-03
ax_lib_sqlite3: fix "unary operator expected"

kvilhaugsvik:


petk:
Thanks @kvilhaugsvik

Probably on today's systems appending x string in if tests is not needed anymore, and only wrapping variable in double quotes should do.. But looks ok to me :+1:

PR #214, opened 2020-02-19
m4/ax_boost_base.m4: Fix debian multiarch_libsubdir path for arch armv7l

pespin:
Without this patch, building on an RPI4 ends up with
BOOST_LDFLAGS=-L/usr/lib while libs are actually under
/usr/lib/arm-linux-gnueabihf, and configure will later file during
AX_BOOST_THREAD macro.

### DIFF
--- a/m4/ax_boost_base.m4
+++ b/m4/ax_boost_base.m4
@@ -33,7 +33,7 @@
 #   and this notice are preserved. This file is offered as-is, without any
 #   warranty.
 
-#serial 47
+#serial 48
 
 # example boost program (need to pass version)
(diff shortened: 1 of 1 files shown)


peti:
Please update the `#serial` number. It must reflect the number of commits that modified the macro. The best solution would be to bump it by 1, commit the change with `--amend`, and then force-push the new version of the PR to github.

MarcoFalke:
@peti Is this ready for merge?

peti:
Yes, it is! Thanks for the ping.

CryptoGodfatherVA4 (commented):
VAR5

PR #219, opened 2020-05-19
ax_gcc_func_attribute.m4: test fails for the fallthrough attribute with clang

CheyenneWills:
The test for the fallthrough attribute fails due to other compiler
warnings when using clang.  The following warnings are generated:

   warning: no case matching constant switch condition '0'
   warning: control reaches end of non-void function [-Wreturn-type]

Because of these additional warnings, the test always results in a
failure for the fallthrough attribute.

Update ax_gcc_func_attribute.m4 to eliminate the warnings that are
masking the test for the fallthrough attribute.

Note, tested the entire suite of attribute tests using clang and only
the fallthrough test contained warnings that were not related to the
the attribute being tested.

### DIFF
--- a/m4/ax_gcc_func_attribute.m4
+++ b/m4/ax_gcc_func_attribute.m4
@@ -78,7 +78,7 @@
 #   and this notice are preserved.  This file is offered as-is, without any
 #   warranty.
 
-#serial 12
+#serial 13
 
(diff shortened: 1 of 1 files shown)


CheyenneWills:
Commit f1d7dd9 just changed the commit message.

PR #221, opened 2020-07-06
Bring back check for default C++ version support.

kvilhaugsvik:
Allow projects that don't care about if the version is with extensions or not to not have the C++ version set unnecessarily.


kvilhaugsvik:
Duplicate of #218

PR #233, opened 2021-06-12
Revert "AX_PTHREAD: target > host"

SoapGentoo:
This reverts commit 2567e0ce0f3a11b535c6b527386197fb49ff172b.

* `AC_CANONICAL_HOST` is the system on which the actual binary will run,
  `AC_CANONICAL_TARGET` is the system for which code is generated. The
  Autoconf manual even mentions that

    `--target=target-type`
      the type of system for which any compiler tools in the package produce code (**rarely needed**).

### DIFF
--- a/m4/ax_pthread.m4
+++ b/m4/ax_pthread.m4
@@ -87,11 +87,11 @@
 #   modified version of the Autoconf Macro, you may extend this special
 #   exception to the GPL to apply to your modified version as well.
 
-#serial 30
+#serial 31
 
 AU_ALIAS([ACX_PTHREAD], [AX_PTHREAD])
 AC_DEFUN([AX_PTHREAD], [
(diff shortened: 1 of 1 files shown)


mattst88:
I think the author @rehsack now agrees with the revert. See https://github.com/autoconf-archive/autoconf-archive/commit/2567e0ce0f3a11b535c6b527386197fb49ff172b#r52114886

@peti: Please merge this. The commit it's reverting has already been upstream since February and is included in the latest tag. We're concerned that downstreams are going to bundle this broken `ax_pthreads.m4` and cause havoc for distributors for ages to come.

mattst88:
Nice, just had to deal with a problem caused by this in https://github.com/fenrus75/powertop/pull/90.

Could you *please* make a release?

PR #239, opened 2021-08-30
AX_CC_MAXOPT: adjust to new way of testing if var is set

jpalus:
autoconf 2.70 changed default value set in $ac_test_* from "set" to "y"
see https://git.savannah.gnu.org/cgit/autoconf.git/commit/?id=76754e0
instead of checking for specific value, just check if var is empty


stevengj:
See also #247 (sorry for the duplicate).  That PR supports both the old style and the new style, in case someone runs this with an older autoconf version.

PR #240, opened 2021-11-05
AX_COMPILER_VENDOR: Add support for the NVIDIA HPC Compiler

chuckatkins:
This specifically detects the NVIDIA HPC compiler.  The "nvhpc"
label is explicitly chosen here insted of "nvidia" to distinguish
from other NVIDIA compilers (cuda, etc.)


zhanhb:
``` txt
syntax error near unexpected token `)'
 `    gnu)'
```

PR #244, opened 2021-12-25
AX_CXX_COMPILE_STDCXX: add C++20 support

manxorist:


manxorist:
Should I explicitly contact previous contributors to this macros for review? <https://www.gnu.org/software/autoconf-archive/How-to-contribute.html> does not talk about that situation.

m4/ax_cxx_compile_stdcxx.m4
```diff
@@ -51,6 +52,7 @@ AC_DEFUN([AX_CXX_COMPILE_STDCXX], [dnl
   m4_if([$1], [11], [ax_cxx_compile_alternatives="11 0x"],
         [$1], [14], [ax_cxx_compile_alternatives="14 1y"],
         [$1], [17], [ax_cxx_compile_alternatives="17 1z"],
+        [$1], [20], [ax_cxx_compile_alternatives="20"],
```
MarcoFalke, line 55:
```suggestion
        [$1], [20], [ax_cxx_compile_alternatives="20 2a"],
```

Missing 2a?

m4/ax_cxx_compile_stdcxx.m4
```diff
@@ -960,3 +973,34 @@ namespace cxx17
...
+#else
+
+#include <version>
+
+namespace cxx20
+{
+
+// As C++20 supports feature test macros in the standard, there is no
+// immediate need to actually test for feature availability on the
+// Autoconf side.
```
MarcoFalke, line 999: I think you'll still need to test for the feature test macros to be available in general? For example via:


```cpp
    namespace language_feature_test {
        static_assert(__cpp_conditional_explicit > 0L);
    }

MarcoFalke (commented):
Missing `m4/ax_cxx_compile_stdcxx_20.m4`? See https://github.com/autoconf-archive/autoconf-archive/pull/245/files#diff-0fedcc4ad33cb9ac406ec509b6168ca917d950349c04f855c701c90cf059d9f5

I didn't realize this pull request existed until I wrote #245. @peti do you have a preference on what approach to take? This one assumes that feature test macros are used, in #245 I checked for the features themselves.

m4/ax_cxx_compile_stdcxx.m4
manxorist: I guess one could do that, however not implemented feature macros are equivalent to all particular feature macro not being announced. So for user code little would change if it checks the feature macros. I guess for code that does not use feature macros in C++17 mode, it might simplify checks.

manxorist:
> Missing `m4/ax_cxx_compile_stdcxx_20.m4`? See https://github.com/autoconf-archive/autoconf-archive/pull/245/files#diff-0fedcc4ad33cb9ac406ec509b6168ca917d950349c04f855c701c90cf059d9f5

Isn't that somewhat redundant? I can of course add it if desired.

> I didn't realize this pull request existed until I wrote #245. @peti do you have a preference on what approach to take? This one assumes that feature test macros are used, in #245 I checked for the features themselves.

I'm not sure either. I myself prefer to have less logic in a build system in general, however that might contradict with autoconf's philosophy.

MarcoFalke (approved):

PR #255, opened 2022-07-24
ax_cxx_compile_stdcxx: Fix for MSVC

ojwb:
MSVC always sets __cplusplus to 199711L in older versions; newer versions
do the same unless /Zc:__cplusplus is specified as well as a `/std:c++<NN>`
switch:

https://devblogs.microsoft.com/cppblog/msvc-now-correctly-reports-__cplusplus/

This fixes a regression compared to the old separate
AX_CXX_COMPILE_STDCXX_11, etc macros which didn't check the value of
__cplusplus so worked OK with MSVC.

### DIFF
--- a/m4/ax_cxx_compile_stdcxx.m4
+++ b/m4/ax_cxx_compile_stdcxx.m4
@@ -43,7 +43,7 @@
 #   and this notice are preserved.  This file is offered as-is, without any
 #   warranty.
 
-#serial 14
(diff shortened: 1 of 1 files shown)


ojwb:
Sorry, realised just after opening the PR that the `#serial` should be bumped too so I did that and force-pushed as it seemed unlikely anyone had started to look at this yet.

PR #261, opened 2022-08-31
AX_PYTHON_DEVEL: fix for Python 3.10+

bastien-roucaries:


bastien-roucaries:
From fedora imported on debian

PR #264, opened 2022-10-30
Clang 16 fixes

thesamesam:
Clang 16 makes -Wimplicit-function-declaration an error by default.

Signed-off-by: Sam James <sam@gentoo.org>


thesamesam:
Ping

PR #265, opened 2022-11-09
Update python macros for PEP 632

bhundven:
PEP 632 deprecates usage of distutils as of python >= 3.10.
Try to use syconfig or packaging instead. Fallback to distutils for legacy
python versions.

Bonus fix:
`ax_python_version` in `AX_PYTHON_CSPEC` didn't properly parse versions like
`python3.10`. It would have returned `3.1`.

Signed-off-by: Bryan Hundven <bryanhundven@gmail.com>

### DIFF
--- a/m4/ax_python.m4
+++ b/m4/ax_python.m4
@@ -50,7 +50,7 @@
 #   modified version of the Autoconf Macro, you may extend this special
 #   exception to the GPL to apply to your modified version as well.
 
-#serial 20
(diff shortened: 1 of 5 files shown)


bhundven:
I'm still testing/fixing a few of these macros. I'm going to convert this to a draft for now.

peti:
Please resubmit when the code is ready!

PR #269, opened 2023-01-06
AX_ADD_FORTIFY_SOURCE: add -D_FORTIFY_SOURCE=3 check

rurban:


thesamesam:
Just to double check: have you tested this with GCC 11 to make sure it correctly detects that it's not supported and uses F_S=2 instead?

thesamesam:
cc @SoapGentoo

rurban:
Yes, I've checked all 3 cases

PR #272, opened 2023-01-19
AX_CXX_COMPILE_STDCXX: Fix MSVC with C++17 or later

ojwb:
MSVC defaults to C++14 and needs -std:c++NN to enable support for newer versions so add that to the list of options to try.

### DIFF
--- a/m4/ax_cxx_compile_stdcxx.m4
+++ b/m4/ax_cxx_compile_stdcxx.m4
@@ -43,7 +43,7 @@
 #   and this notice are preserved.  This file is offered as-is, without any
 #   warranty.
 
-#serial 15
+#serial 16
 
 dnl  This macro is based on the code from the AX_CXX_COMPILE_STDCXX_11 macro
 dnl  (serial version number 13).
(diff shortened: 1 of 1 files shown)


ojwb:
Sorry, there's a problem with patch - `-std:c++17` gets mapped to the same cache variable as `-std=c++17` (because `:` and `=` are both flattened to `_`).  Looking into how best to solve this, but please don't merge this as-is.

c72578:
@ojwb In the meantime you can convert this PR to draft.

[Converting a pull request to a draft](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/changing-the-stage-of-a-pull-request#converting-a-pull-request-to-a-draft)

ojwb:
I've pushed a fix.

I've gone for just suffixing the cache variables for the `-std:c++NN` options with `_MSVC` which seems unlikely to accidentally collide with something else, but doesn't seem terribly elegant.

MSVC would also accept `/std:c++17`, etc but `/` and `-` both get folded to `_` so that doesn't help.  FWIW I chose the `-` variant as other compilers should simply reject it as an unknown option rather than trying to open a file in the root of the filing system.

PR #275, opened 2023-04-10
ax_cc_tentdef: test for C tentative definitions

cstes:
Hi, 

I wrote this macro more than 25 years ago but it is of renewed interest due to 
https://gcc.gnu.org/gcc-10/portint_to.html

The test checks whether the C compiler somehow supports "C tentative definitions" as discussed in the K&R book on C.

I submitted this a wihle ago on GNU savannah (patch #10317) but received no feedback so I'll submit it here to see whether this is the appropriate channel for submissions ...

Regards,
David Stes

### DIFF
--- a/m4/ax_cc_attrcommon.m4
+++ b/m4/ax_cc_attrcommon.m4
@@ -1,0 +1,76 @@
+# ===========================================================================
+#       https://www.gnu.org/software/autoconf-archive
+# ===========================================================================
+#
+# SYNOPSIS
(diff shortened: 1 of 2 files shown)


cstes:
correct link for gcc-10 is

https://gcc.gnu.org/gcc-10/porting_to.html

The test itself is not specific at all to GCC and has been tested with many different C compilers both on UNIX and non UNIX platforms.

PR #280, opened 2023-06-07
ax_blas: detect flexiblas

theAeon:


peti (changes requested):
Please explain the motivation for this change in more detail. What is flexiblas and why is it needed? Why is it needed only on some platforms?

theAeon:
Sure. Flexiblas is a wrapper for the various supported blas implementations. Its also the default `-lblas` library on every fedora since 33 and EL9 equivalents. That's a significant number of machines to leave out.

PR #284, opened 2023-08-12
Make macros `-Wstrict-prototypes` compatible

SoapGentoo:
* With Clang having switched to modern C conventions already and GCC following suit soon, we want to have macros that work in the presence of `-Wstrict-prototypes` in all C language modes, not just C23.


SoapGentoo:
@peti ping

PR #287, opened 2023-09-25
Add Lua 5.4 support to AX_PROG_LUA macro

alerque:
Since the current Lua version is 5.4, many distros package it with a bare `lua` binary and suffix older builds with the version number. This is *not* universally true however, nor is it forward-compatible. On systems where there is either no default Lua or the default Lua is set to an older version, searching for the binary with the version suffix is the only way to find the newest one. This should be done in priority to falling back to older versions for consistency.

### DIFF
--- a/m4/ax_lua.m4
+++ b/m4/ax_lua.m4
@@ -19,7 +19,7 @@
 #   header is checked to match the Lua interpreter version exactly. When
 #   searching for Lua libraries, the version number is used as a suffix.
 #   This is done with the goal of supporting multiple Lua installs (5.1,
-#   5.2, and 5.3 side-by-side).
+#   5.2, 5.3, and 5.4 side-by-side).
 #
 #   A note on compatibility with previous versions: This file has been
 #   mostly rewritten for serial 18. Most developers should be able to use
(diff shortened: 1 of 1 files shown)


cooljeanius (approved):

m4/ax_lua.m4
```diff
@@ -181,7 +181,7 @@
 #   modified version of the Autoconf Macro, you may extend this special
 #   exception to the GPL to apply to your modified version as well.
 
-#serial 42
+#serial 43
```
peti, line 184: The serial number is the number of commits that modify this macro. This means that every commit must bump it. Otherwise the test suite fails.

peti (changes requested):
Please make sure the test suite succeeds after these changes.

m4/ax_lua.m4
alerque: Fair enough. I wasn't sure how PR history was going to be handled, but this is two distinct operations, one bug fix and one feature, so I did two commits and now have made it two serials.

I'll have a much more extensive addition with LuaJIT support in another PR soon.

alerque: The test suite doesn't pass for me locally, but all the errors are about *other* macros untouched by this PR. It's hard for me to tell what's going on remotely since the test suite doesn't run in CI when I push (presumably first-time contributor PRs are blocked from automatic CI jobs).

peti (approved):

PR #288, opened 2023-09-26
Add LuaJIT support to AX_PROG_LUA macro

alerque:
This is a follow up to #287.

This is a much bigger change adding support for LuaJIT as a Lua interpreter. This is a tricky addition because two sets of version numbers are involved: the Lua VM version (which for LuaJIT is *lower* that current PUC Lua releases, 5.1 vs 5.4) and also a JIT version. It is also quite common to have both PUC Lua and LuaJIT variants installed on a system, but the lower numbering scheme does not mean that PUC should be preferred.

This implementation leaves it up to the user to specifically request LuaJIT be considered by adding a `--with-luajit` configure option. If they do that and it is set to true, then the regular Lua macros will function in the usual way but with the binary names, include paths, and other things tweaked for use with LuaJIT.

I've been using [earlier iterations](https://github.com/sile-typesetter/sile/blob/master/build-aux/ax_lua.m4) of this (based on a much earlier serial number) for several years with SILE on [a number of platforms](https://repology.org/project/sile/versions) including Arch Linux, Ubuntu, Nix, Homebrew, OpenBSD, and many others. It has gone pretty well so far. This version is a bit cleaned up and rebased on serial 44.

### DIFF
--- a/m4/ax_lua.m4
+++ b/m4/ax_lua.m4
@@ -49,6 +49,14 @@
 #   interpreter. If LUA is blank, the user's path is searched for an
 #   suitable interpreter.
 #
+#   Optionally a LUAJIT option may be set ahead of time to look for and
+#   validate a LuaJIT install instead of PUC Lua. Usage might look like:
+#
+#     AC_ARG_WITH(luajit, [AS_HELP_STRING([--with-luajit],
+#         [Prefer LuaJIT over PUC Lua, even if the latter is newer. Default: no])
+#     ])
+#     AM_CONDITIONAL([LUAJIT], [test "x$with_luajit" != 'xno'])
+#
 #   If MINIMUM-VERSION is supplied, then only Lua interpreters with a
 #   version number greater or equal to MINIMUM-VERSION will be accepted. If
 #   TOO-BIG-VERSION is also supplied, then only Lua interpreters with a
@@ -152,6 +160,7 @@
 #
 # LICENSE
 #
+#   Copyright (c) 2023 Caleb Maclennan <caleb@alerque.com>
 #   Copyright (c) 2015 Reuben Thomas <rrt@sc3d.org>
 #   Copyright (c) 2014 Tim Perkins <tprk77@gmail.com>
 #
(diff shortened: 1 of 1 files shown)


m4/ax_lua.m4
```diff
@@ -49,6 +49,14 @@
 #   interpreter. If LUA is blank, the user's path is searched for an
 #   suitable interpreter.
 #
+#   Optionally a LUAJIT option may be set ahead of time to look for and
+#   validate a LuaJIT install instead of PUC Lua. Usage might look like:
+#
+#     AC_ARG_WITH(luajit, [AS_HELP_STRING([--with-luajit],
+#         [Prefer LuaJIT over PUC Lua, even if the latter is newer. Default: no])
+#     ])
+#     AM_CONDITIONAL([LUAJIT], [test "x$with_luajit" != 'xno'])
```
eli-schwartz, line 58: > This implementation leaves it up to the user to specifically request LuaJIT be considered by adding a `--with-luajit` configure option. If they do that and it is set to true, then the regular Lua macros will function in the usual way but with the binary names, include paths, and other things tweaked for use with LuaJIT.

This doesn't seem to be very optional at all: https://bugs.gentoo.org/941718

Per the automake documentation:

> Before using a conditional, you must define it by using AM_CONDITIONAL in the configure.ac file (see [Autoconf macros supplied with Automake](https://www.gnu.org/software/automake/manual/html_node/Macros.html)).

So any previously existing users of ax_lua.m4 now get a fatal autoreconf error if they don't support luajit, which appears to be precisely the opposite of what you wanted. :(

```
configure.ac:641: error: AM_COND_IF: no such condition "LUAJIT"
/usr/share/aclocal-1.17/cond-if.m4:23: AM_COND_IF is expanded from...
/usr/share/aclocal/ax_lua.m4:199: AX_PROG_LUA is expanded from...
```

alerque: Yes, my bad. All my test projects had this conditional set one way or the other. I tested with default on and default off and using config flag overrides both ways, but apparently not without doing anything at all. See #313 for a fix.

PR #290, opened 2023-11-21
ax_python_devel: fix typo preventing correct variable evaluation

mhei:
The test should check the variable, not the fixed string. Most probably a simple typo, so fix it.


petk (approved):

petk:
Thank you @mhei

PR #292, opened 2023-12-25
Fix typos found by codespell

DimitriPapadopoulos:
* Both `re-usable` and `reusable` can be seen as correct, but:
  * Majors dictionaries suggest `reuse` only:
    * [OED](https://www.oed.com/search/dictionary/?q=re-use)
    * [Cambridge](https://www.oxfordlearnersdictionaries.com/definition/english/reuse_1)
    * [Collins](https://www.collinsdictionary.com/dictionary/english/reuse)
    * [Longman](https://www.ldoceonline.com/dictionary/reuse)
    * [Merriam-Webster](https://www.merriam-webster.com/dictionary/reuse)
    * [Oxford Learner's Dictionary](https://dictionary.cambridge.org/dictionary/english/reuse)
  * From [The Blue Book of Grammar and Punctuation](https://www.grammarbook.com/blog/hyphens/hyphens-with-the-prefix-re/):
    > Use the hyphen with the prefix _re_ only when _re_ means _again_ AND omitting the hyphen would cause confusion with another word.
* Both `tesselation` and `tessellation` can be seen as correct, but:
  * The [Google Ngram Viewer](https://books.google.com/ngrams/graph?content=tesselation%2Ctessellation&year_start=1800&year_end=2019&corpus=en-2019) shows `tessellation`  has been in much wider use.
  * The [GLU Library](http://archive.irixnet.org/siliconsurf/tech/openGL/mjk.intro/subsection3_2_6.html) documentation uses `tessellation`

### DIFF
--- a/m4/ax_cpu_freq.m4
+++ b/m4/ax_cpu_freq.m4
@@ -19,7 +19,7 @@
 #   and this notice are preserved. This file is offered as-is, without any
 #   warranty.
 
-#serial 10
+#serial 11
 
 AC_DEFUN([AX_CPU_FREQ],
 [AC_REQUIRE([AC_PROG_CC])
@@ -88,7 +88,7 @@ static float average_MHz(int tries = 2)
     of.close()
 ])],
      [ax_cpu_freq=`cat conftest_cpufreq`; rm -f conftest_cpufreq],
-     [ax_cpu_freq=unknow; rm -f conftest_cpufreq]
+     [ax_cpu_freq=unknown; rm -f conftest_cpufreq]
  )])
 AC_LANG_POP([C++])
 
--- a/doc/autoconf-archive.texi
+++ b/doc/autoconf-archive.texi
@@ -11,7 +11,7 @@
 
 @copying
 This manual is for GNU Autoconf Archive version @value{VERSION}, a collection of
-freely re-usable Autoconf macros.
+freely reusable Autoconf macros.
 
 Copyright @copyright{} 2022 Autoconf Archive Maintainers
 @email{autoconf-archive-maintainers@@gnu.org}
@@ -26,7 +26,7 @@ Documentation License''.
 
 @dircategory Software development
 @direntry
-* Autoconf Archive: (autoconf-archive). A collection of re-usable Autoconf macros.
+* Autoconf Archive: (autoconf-archive). A collection of reusable Autoconf macros.
 @end direntry
 
 @titlepage
--- a/m4/ax_check_glu.m4
+++ b/m4/ax_check_glu.m4
@@ -49,7 +49,7 @@
 #   framework.
 #
 #   Some implementations (in particular, some versions of Mac OS X) are
-#   known to treat the GLU tesselator callback function type as "GLvoid
+#   known to treat the GLU tessellator callback function type as "GLvoid
 #   (*)(...)" rather than the standard "GLvoid (*)()". If the former
 #   condition is detected, this macro defines "HAVE_VARARGS_GLU_TESSCB".
 #
@@ -85,7 +85,7 @@
 #   modified version of the Autoconf Macro, you may extend this special
 #   exception to the GPL to apply to your modified version as well.
 
-#serial 23
+#serial 24
 
 # example program
 m4_define([_AX_CHECK_GLU_PROGRAM],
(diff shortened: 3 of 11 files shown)


m4/ax_cpu_freq.m4
```diff
@@ -88,7 +88,7 @@ static float average_MHz(int tries = 2)
     of.close()
 ])],
      [ax_cpu_freq=`cat conftest_cpufreq`; rm -f conftest_cpufreq],
-     [ax_cpu_freq=unknow; rm -f conftest_cpufreq]
+     [ax_cpu_freq=unknown; rm -f conftest_cpufreq]
```
DimitriPapadopoulos, line 91: I hope this one was not on purpose.

petk: I think it's ok to change it. This AC_DEFINE should probably also be a quoted string. But this can be opened in a separate PR.

petk (approved):

petk:
Ah, I see, this last fail in the CI was only related to separate commit itself since the CI checks commits separately. Otherwise all looks ok here.

DimitriPapadopoulos:
Squashed into a single commit.

spaette:
ax_cpu_freq.m4

personally I wouldn't modify the above file

ax_dist_msi.m4

typo `addind` on line 56

ax_prog_emacs.m4

typo `failurefully` on line 25

ax_sys_perlsharpbang.m4

possibly change this comment line

`# bec it must be absent in Makefile`

to

`# as it must not appear in Makefile`

DimitriPapadopoulos:
> ax_cpu_freq.m4
> ```diff
> -     [ax_cpu_freq=unknow; rm -f conftest_cpufreq]
> +     [ax_cpu_freq=unknown; rm -f conftest_cpufreq]
> ```
> > Was this one on purpose?
> 
> Is the string an action or the result of something?
> 
> such clarification might facilitate a decision on whether or not to modify the string

It's clearly a result, that is later used to define `CPU_FREQ`. Two issues here:
- backwards compatibility,
- the possibility that the result is misspelled on purpose — but I cannot imagine why.

The misspelling has been there from the start (https://github.com/autoconf-archive/autoconf-archive/commit/6c640dfe1e1a4cc53400a24992c058525dd69033 in 2007).

> ax_prog_emacs.m4
> 
> typo `failurefully` on line 25

I could change _exit failurefully_ into _exit in error_. However, wasn't that an intentional pun?

spaette:
```
$ sed -n '20,24p' /tmp/autoconf-archive/m4/ax_gcc_x86_avx_xgetbv.m4
#   If the xgetbv instruction fails (because you are running a
#   cross-compiler, or because you are not using gcc, or because you are on
#   a processor that doesn't have this instruction),
#   ax_cv_gcc_x86_avx_xgetbv_OP is set to the string "unknown".
#
$ sed -n '75,77p' /tmp/autoconf-archive/m4/ax_gcc_x86_avx_xgetbv.m4
     [ax_cv_gcc_x86_avx_xgetbv_$1=`cat conftest_xgetbv`; rm -f conftest_xgetbv],
     [ax_cv_gcc_x86_avx_xgetbv_$1=unknown; rm -f conftest_xgetbv],
     [ax_cv_gcc_x86_avx_xgetbv_$1=unknown])])
$ sed -n '22,25p' /tmp/autoconf-archive/m4/ax_gcc_x86_cpuid.m4
#   If the cpuid instruction fails (because you are running a
#   cross-compiler, or because you are not using gcc, or because you are on
#   a processor that doesn't have this instruction), ax_cv_gcc_x86_cpuid_OP
#   is set to the string "unknown".
$ sed -n '85,87p' /tmp/autoconf-archive/m4/ax_gcc_x86_cpuid.m4
     [ax_cv_gcc_x86_cpuid_$1=`cat conftest_cpuid`; rm -f conftest_cpuid],
     [ax_cv_gcc_x86_cpuid_$1=unknown; rm -f conftest_cpuid],
     [ax_cv_gcc_x86_cpuid_$1=unknown])])
$ 
```

spaette:
@tnguyen-rh

Was the intention `exit failure fully`?

```
$ sed -n '21,25p' /tmp/autoconf-archive/m4/ax_prog_emacs.m4
#   If env var EMACS is set, try to use its value directly, but avoid
#   getting fooled by value 't' (set by older Emacsen for subprocesses). If
#   no joy from the environment, search for "emacs" via AC_CHECK_PROG. If
#   still no joy, display "Emacs not found; required!" and make configure
#   exit failurefully. Otherwise, set shell var EMACS and AC_SUBST it, too.
$ 
```

spaette:
@stevengj

I guess you are one of the listed authors of ax_gcc_x86_avx_xgetbv.m4.

Would you have additional explanation of the code in ax_cpu_freq.m4?

PR #294, opened 2024-03-31
m4_define with braces around concat source, m4_if without extra arg

ggraham:
m4_if had an extra [] in the params
m4_define was not define-ing without [...] around the source to be concat'd

unsure if the m4_if extra param was a holdover, or if ext|noext was not used frequently enough to be noticed
unsure if prior m4_define dealt with source not between [...] but this seems to resolve the issue

### DIFF
--- a/m4/ax_cxx_compile_stdcxx.m4
+++ b/m4/ax_cxx_compile_stdcxx.m4
@@ -75,7 +75,7 @@ AC_DEFUN([AX_CXX_COMPILE_STDCXX], [dnl
       ac_success=yes
     fi])
 
-  m4_if([$2], [noext], [], [dnl
+  m4_if([$2], [noext], [dnl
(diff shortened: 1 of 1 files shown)


manxorist:
@ggraham 
@peti

> m4_if had an extra [] in the params

This was **NOT** broken, You broke it by inverting the intended logic.

`AX_CXX_COMPILE_STDCXX(20, [], [optional])` now even fails to enable ANY C++20 support at all, and ext and noext cause inverted results.

Please at the very least test your changes.

ext|noext **IS** used frequently, and it would have been noticed if it was actually broken.

As autoconf-archive takes MONTHS to apply changes and then again MONTHS for a release, and AGAIN months for distributions to catch up, the thing you just broke will haunt us probably for years.

See #307 for the revert.

Yes, I am angry.

PR #297, opened 2024-05-13
Fix AC_CACHE_VAL warning

petk:
Cache variables must contain `_cv_` to be cached otherwise this warning is emitted when generating configure script:

    AC_CACHE_VAL(ax_cpu_freq, ...): suspicious cache-id, must contain _cv_
    to be cached

I'll wait with this one for the existing PR #292 to be merged first

### DIFF
--- a/m4/ax_cpu_freq.m4
+++ b/m4/ax_cpu_freq.m4
@@ -19,12 +19,12 @@
 #   and this notice are preserved. This file is offered as-is, without any
 #   warranty.
 
-#serial 11
+#serial 12
 
 AC_DEFUN([AX_CPU_FREQ],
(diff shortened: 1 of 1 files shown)


peti:
I believe this PR is ready to be merged now?

petk:
Probably more fine tuning could be done here but I wouldn't add more to this PR for now. For example, if the value is "unknown" then the AC_DEFINE_UNQUOTED shouldn't be defined. And 3rd argument needs to be added to AC_RUN_IFELSE (for cross-compiling)...

petk:
I've appended all mentioned fixes in this PR as the check would not be properly functional otherwise. Details in the commit message.

PR #298, opened 2024-05-23
ax_code_coverage: remove /tmp/* from remove patterns failing with lcov 2.0

janzizazizka:
The lcov >= 2.0 refactored error handling and by default it fails when an unused remove pattern is provided:

  LCOV   --remove /tmp/* */tests/*.c
lcov: ERROR: 'exclude' pattern '/tmp/*' is unused.
        (use "lcov --ignore-errors unused ..." to bypass this error)

Based on the macro history is is not clear why the /tmp/* would be needed by default. Usual autotools projects don't use /tmp/. If needed users can still add this pattern to
CODE_COVERAGE_IGNORE_PATTERN.

This can be reproduced for example in Fedora 40 which comes with lcov 2.0:

```
docker run --rm -it fedora:40                                                  
                                                                               
dnf install -y gcc automake autoconf gawk lcov gcovr git autoconf-archive      
                                                                               
git clone https://github.com/janzizazizka/autotools-template.git               
cd autotools-template                                                          
git checkout dd8ecefd9de42305e9b7914c6b3bd01d2c0a3021                          
                                                                               
./autogen.sh                                                                   
./configure --enable-code-coverage                                             
make check-code-coverage
```

Results with:

```
  LCOV   --remove /tmp/* */tests/*.c
lcov: ERROR: 'exclude' pattern '/tmp/*' is unused.
        (use "lcov --ignore-errors unused ..." to bypass this error)
```

Or does anyone know why we should have `/tmp/*` by default in remove pattern?

### DIFF
--- a/m4/ax_code_coverage.m4
+++ b/m4/ax_code_coverage.m4
@@ -144,7 +144,7 @@ code_coverage_v_lcov_cap_ = \$(code_coverage_v_lcov_cap_\$(AM_DEFAULT_VERBOSITY)
 code_coverage_v_lcov_cap_0 = @echo \"  LCOV   --capture\" \$(CODE_COVERAGE_OUTPUT_FILE);
 code_coverage_v_lcov_ign = \$(code_coverage_v_lcov_ign_\$(V))
 code_coverage_v_lcov_ign_ = \$(code_coverage_v_lcov_ign_\$(AM_DEFAULT_VERBOSITY))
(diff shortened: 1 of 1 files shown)


evjeesm:
After my github-actions runner was updated to ubuntu 24.04.1 (latest) it received new version of lcov.
This also broke my CI/CD workflow. I hope maintainers will address this issue!
For now I will downgrade my runner to ubuntu 22.04.

PR #300, opened 2024-05-25
AX_CXX_COMPILE_STDCXX changes

ojwb:
There are three changes here, each as a separate commit.  They are related so it seemed simplest to just open one PR for them, but if you'd prefer I can open a series of PRs.

With these changes we seem to correctly probe C++17, C++20 and C++23 support with various versions of GCC, clang and MSVC.

### DIFF
--- a/m4/ax_cxx_compile_stdcxx.m4
+++ b/m4/ax_cxx_compile_stdcxx.m4
@@ -43,7 +43,7 @@
 #   and this notice are preserved.  This file is offered as-is, without any
 #   warranty.
 
-#serial 18
+#serial 19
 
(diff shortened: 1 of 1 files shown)


ojwb:
It seems the tooling expects the serial to get bumped for every git commit which touches the file, so I've adjusted the branch to do that and force pushed, and now CI passes.

manxorist:
Sadly this has been merged with a totally broken commit on top of it, see <https://github.com/autoconf-archive/autoconf-archive/commit/5b0d43b767bf3f800f97892905b6d1ba50150f1a#commitcomment-147999469>.

PR #301, opened 2024-05-30
ax_check_compile_flag: add -Werror when checking for compiler flags

zyv:
Seemingly that at some point clang started accepting _any_ flags, whereas previously it would error out.

These days, you can give it -Whamsandwich and it will succeed, while at the same time throwing an annoying warning.

Add -Werror so that everything gets flagged and set accordingly.

Bugzilla: https://bugs.freedesktop.org/show_bug.cgi?id=108082
Cc: Vinson Lee <vlee@freedesktop.org>
Repored-by: Vinson Lee <vlee@freedesktop.org>

Reviewed-by: Dylan Baker <dylan@pnwbakers.com>
Reviewed-by: Eric Engestrom <eric.engestrom@intel.com>

### DIFF
--- a/m4/ax_check_compile_flag.m4
+++ b/m4/ax_check_compile_flag.m4
@@ -41,7 +41,14 @@ AC_DEFUN([AX_CHECK_COMPILE_FLAG],
 AS_VAR_PUSHDEF([CACHEVAR],[ax_cv_check_[]_AC_LANG_ABBREV[]flags_$4_$1])dnl
 AC_CACHE_CHECK([whether the _AC_LANG compiler accepts $1], CACHEVAR, [
   ax_check_save_flags=$[]_AC_LANG_PREFIX[]FLAGS
-  if test x"$GCC" = xyes -o x"$GFC" = xyes -o x"$G77" = xyes -o x"$GOBJC" = xyes -o x"$GOBJCXX" = xyes ; then
+  if test x"m4_case(_AC_LANG,
+                     [C], [$GCC],
+                     [C++], [$GXX],
+                     [Fortran], [$GFC],
+                     [Fortran 77], [$G77],
+                     [Objective C], [$GOBJC],
+                     [Objective C++], [$GOBJCXX],
+                     [no])" = xyes ; then
     add_gnu_werror="-Werror"
   fi
   _AC_LANG_PREFIX[]FLAGS="$[]_AC_LANG_PREFIX[]FLAGS $4 $1 $add_gnu_werror"
--- a/m4/ax_check_compile_flag.m4
+++ b/m4/ax_check_compile_flag.m4
@@ -34,14 +34,14 @@
 #   and this notice are preserved.  This file is offered as-is, without any
 #   warranty.
 
-#serial 7
+#serial 8
 
 AC_DEFUN([AX_CHECK_COMPILE_FLAG],
 [AC_PREREQ(2.64)dnl for _AC_LANG_PREFIX and AS_VAR_IF
 AS_VAR_PUSHDEF([CACHEVAR],[ax_cv_check_[]_AC_LANG_ABBREV[]flags_$4_$1])dnl
 AC_CACHE_CHECK([whether the _AC_LANG compiler accepts $1], CACHEVAR, [
   ax_check_save_flags=$[]_AC_LANG_PREFIX[]FLAGS
-  _AC_LANG_PREFIX[]FLAGS="$[]_AC_LANG_PREFIX[]FLAGS $4 $1"
+  _AC_LANG_PREFIX[]FLAGS="$[]_AC_LANG_PREFIX[]FLAGS $4 $1 -Werror"
   AC_COMPILE_IFELSE([m4_default([$5],[AC_LANG_PROGRAM()])],
     [AS_VAR_SET(CACHEVAR,[yes])],
     [AS_VAR_SET(CACHEVAR,[no])])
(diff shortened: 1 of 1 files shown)


ojwb:
The change proposed here will likely cause problems with a compiler which doesn't support `-Werror` since the macro will think any probed option is unsupported if `-Werror` itself result in an "unknown option" or similar error.

E.g. https://godbolt.org/z/14YhG7T6K shows MSVC with `-Werror` gives:

```
    Output of x64 msvc v19.40 VS17.10 (Compiler #1)

cl : Command line error D8021 : invalid numeric argument '/Werror'
Compiler returned: 2
```

I think you'd need to probe whether `-Werror` itself is a supported compiler option rather than unconditionally using it.

(This is a very long standing feature of clang BTW - the oldest version compiler explorer seems to have is clang 3.0.0 which gives the same warning as current clang versions do:

```
warning: unknown warning option '-Whamsandwich' [-Wunknown-warning-option]
1 warning generated.
Compiler returned: 0
```
)

zyv:
I didn't know supporting MSVC would be even a factor in the context of autoconf.

If you want `-Werror` to be conditionally added though, the macro needs a major rewrite, I'm afraid. First we need to check for `-Werror` and then check for the requested flag either with `-Werror` or not.

Is this going to be accepted?

ojwb:
> I didn't know supporting MSVC would be even a factor in the context of autoconf.

autotools does support MSVC, and also other compilers beyond those that support `-Werror` (AFAIK that's only GCC, clang and maybe Intel's compiler).  MSVC was just an easy one to demonstrate because it's in compiler explorer.

> Is this going to be accepted?

I'm not the maintainer so I can't answer that.

I just thought I should flag up a problem I'd noticed as the maintainer seems to be busy currently.

ojwb:
Another approach occurred to me - `AC_PROG_CC` is documented to set shell variable `GCC` to `yes` and it just checks the compiler defines `__GNUC__` so this will also be true for clang (and any other compilers which masquerade as GCC - I think Intel's icc does too) so this should be a safe basis for when to use `-Werror`.

The slightly fiddly bit is the name of the shell variable to check depends on the language - `GCC` is only for C, for C++ it's `GXX`, for Fortran it's `GFC`, for Objective C it's `GOBJC`, and so on.  Searching for `If using a compiler that supports GNU` in the autoconf manual seems to work to find them.

zyv:
I have added the conditional as you suggested, thanks @ojwb .

/cc @peti any opinion on that?

m4/ax_check_compile_flag.m4
```diff
@@ -41,7 +41,10 @@ AC_DEFUN([AX_CHECK_COMPILE_FLAG],
 AS_VAR_PUSHDEF([CACHEVAR],[ax_cv_check_[]_AC_LANG_ABBREV[]flags_$4_$1])dnl
 AC_CACHE_CHECK([whether the _AC_LANG compiler accepts $1], CACHEVAR, [
   ax_check_save_flags=$[]_AC_LANG_PREFIX[]FLAGS
-  _AC_LANG_PREFIX[]FLAGS="$[]_AC_LANG_PREFIX[]FLAGS $4 $1 -Werror"
+  if test x"$GCC" = xyes -o x"$GFC" = xyes -o x"$G77" = xyes -o x"$GOBJC" = xyes -o x"$GOBJCXX" = xyes ; then
+    add_gnu_werror="-Werror"
+  fi
+  _AC_LANG_PREFIX[]FLAGS="$[]_AC_LANG_PREFIX[]FLAGS $4 $1 $add_gnu_werror"
```
ojwb, line 47: You're missing `$GXX` for C++.

Also the autoconf manual advises to avoid `test -o` and instead use `||`, i.e. `test x"$GCC" = xyes ||test x"$GFC" = xyes` etc.

However you'll add `-Werror` if **any** of the probed compilers is detected as the GNU one, so if you're using a GNU compiler for one language and a non-GNU one for another, both will get `-Werror` but the non-GNU one may not support it.  That's certainly a corner case, but these macros aim to be reusable so should try to cope with unusual situations.

I think it should just check the variable for the current language - here's an implementation which picks the appropriate variable to check at m4 time (with a fall-back to not adding `-Werror` for languages where there's no GNU compiler):

```
if test x"m4_case(_AC_LANG,
                   [C], [$GCC],
                   [C++], [$GXX],
                   [Fortran], [$GFC],
                   [Fortran 77], [$G77],
                   [Objective C], [$GOBJC],
                   [Objective C++], [$GOBJCXX],
                   [no])" = xyes ; then
  add gnu_werror="-Werror"
fi
```

(Lightly tested - I manually checked the generated code for C, C++ and the fall-back case.)

zyv: Thank you for your educative comments.

> You're missing $GXX for C++.

Good point, not sure how that happened, I looked through the manual and maybe there was no comment about the variable... or I just missed it.

> Also the autoconf manual advises to avoid `test -o`...

Well, it's all over the place in the archive, I just looked at another macro and copied & pasted from there. Maybe it makes sense to do a mass replacement, otherwise it gets really confusing.

> (Lightly tested - I manually checked the generated code for C, C++ and the fall-back case.)

Works for me, committed a fixup.

MisterDA:
I confirm that using MSVC and clang-cl with Autotools is a workflow we use in the OCaml compiler. MSVC uses `-WX` to turn warnings into errors, and since Visual Studio 2022 version 17.0 Preview 4 uses `-options:strict` to error on unrecognized options. Clang-cl recognizes `-WX` but not `-options:strict` (see https://github.com/llvm/llvm-project/issues/92911).

zyv:
> I confirm that using MSVC and clang-cl with Autotools is a workflow we use in the OCaml compiler. MSVC uses `-WX` to turn warnings into errors, and since Visual Studio 2022 version 17.0 Preview 4 uses `-options:strict` to error on unrecognized options. Clang-cl recognizes `-WX` but not `-options:strict` (see [llvm/llvm-project#92911](https://github.com/llvm/llvm-project/issues/92911)).

Feel free to submit a PR to add a branch to support MSVC. I have no personal interest in working on it... unless it's going to be bankrolled by Microsoft 😉

PR #307, opened 2024-10-16
Revert broken part of commit 5b0d43b767bf3f800f97892905b6d1ba50150f1a

manxorist:
This commit breaks AX_CXX_COMPILE_STDCXX noext/ext option handling, which caused
"AX_CXX_COMPILE_STDCXX(20, [noext], [optional])"
to result in
"checking whether g++ supports C++20 features with -std=gnu++20... yes"
which is not the desired outcome when using "noext". It should check for "-std=c++20" instead for "-std=gnu++20", as it did before.

See
<https://github.com/autoconf-archive/autoconf-archive/pull/300#issuecomment-2416443298> and
<https://github.com/autoconf-archive/autoconf-archive/commit/5b0d43b767bf3f800f97892905b6d1ba50150f1a#commitcomment-147999469> .

### DIFF
--- a/m4/ax_cxx_compile_stdcxx.m4
+++ b/m4/ax_cxx_compile_stdcxx.m4
@@ -36,7 +36,7 @@
 #   Copyright (c) 2016, 2018 Krzesimir Nowak <qdlacz@gmail.com>
 #   Copyright (c) 2019 Enji Cooper <yaneurabeya@gmail.com>
 #   Copyright (c) 2020 Jason Merrill <jason@redhat.com>
-#   Copyright (c) 2021 Jörn Heusipp <osmanx@problemloesungsmaschine.de>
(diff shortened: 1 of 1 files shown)


manxorist:
See also <https://github.com/autoconf-archive/autoconf-archive/pull/294#issuecomment-2417233165>

manxorist:
Thanks for merging this quickly!

whitslack:
This fix is bad. Now, if you specify neither `ext` nor `noext`, then `AX_CXX_COMPILE_STDCXX` will not check for _any_ `-std=` flags whatsoever, and if your compiler does not support the requested standard by default, then the test fails.

The [documentation](https://www.gnu.org/software/autoconf-archive/ax_cxx_compile_stdcxx.html) says: “If neither is specified, you get whatever works, with preference for no added switch, and then for an extended mode.” So specifying neither `ext` nor `noext` is supposed to fall back to `ext` if the compiler does not support the desired mode by default.

PR #308, opened 2024-10-16
ax_cxx_compile_stdcxx: Fix consecutive testing of older standard versions.

manxorist:
Testing with a compiler that supports C++20 but not C++23, the following snippet
```
AX_CXX_COMPILE_STDCXX(23, [noext], [optional])
AS_IF([test "x$HAVE_CXX23" != "x1"],
 AX_CXX_COMPILE_STDCXX(20, [noext], [optional])
 AS_IF([test "x$HAVE_CXX20" != "x1"],
  [
   AX_CXX_COMPILE_STDCXX(17, [noext], [mandatory])
  ],[]
 )
)
```
results in
```
checking whether g++ supports C++23 features with -std=c++23... no
checking whether g++ supports C++23 features with +std=c++23... no
checking whether g++ supports C++23 features with -h std=c++23... no
checking whether g++ supports C++23 features with -std:c++23... no
configure: No compiler with C++23 support was found
checking whether g++ supports C++20 features with -std=c++20... no
checking whether g++ supports C++20 features with +std=c++20... no
checking whether g++ supports C++20 features with -h std=c++20... no
checking whether g++ supports C++20 features with -std:c++20... no
configure: No compiler with C++20 support was found
checking whether g++ supports C++17 features with -std=c++17... no
checking whether g++ supports C++17 features with +std=c++17... no
checking whether g++ supports C++17 features with -h std=c++17... no
checking whether g++ supports C++17 features with -std:c++17... no
checking whether g++ supports C++17 features with -std=c++1z... no
checking whether g++ supports C++17 features with +std=c++1z... no
checking whether g++ supports C++17 features with -h std=c++1z... no
checking whether g++ supports C++17 features with -std:c++1z... no
configure: error: *** A compiler with support for C++17 language features is required.
```
, which is obviously not what was intended.

The intended result is

```
checking whether g++ supports C++23 features with -std=c++23... no
checking whether g++ supports C++23 features with +std=c++23... no
checking whether g++ supports C++23 features with -h std=c++23... no
checking whether g++ supports C++23 features with -std:c++23... no
configure: No compiler with C++23 support was found
checking whether g++ supports C++20 features with -std=c++20... yes
```
, which this pull request provides.

I am unsure why exactly this happens as I do not consider myself an Autoconf or m4 or shell expert. A little bit of printf-style debugging revealed that it looks like something in Autoconf caches the `AC_COMPILE_IFELSE` invocations for the individual `_AX_CXX_COMPILE_STDCXX_testbody_new_in_??` code blocks across invocations, even with different compiler flags (i.e. `_AX_CXX_COMPILE_STDCXX_testbody_new_in_20` fails to compile with `-std=c++23` because the compiler does not support C++23, and then, when checking for C++20, it does not get re-checked with `-std=c++20`). So, there might be a more Autoconf-correct-ish way to "fix" this.

However, as C++ standards are generally supersets of one another (and in cases where they are not, the user should expect that features which are removed in newer standard are actually not available when choosing such a standard version), only checking for features and/or the monotonically increasing `__cplusplus` value of the precise standard version requested, should be enough and yield identical results in all cases. It should also make the checks slightly faster because less work is done (I did not measure).

I am leaving this as a Draft right now as I would prefer some input from more Autoconf-savy people (highlighting anyone who has contributed to `ax_cxx_compile_stdcxx.m4` since roughly C++17: @krnowak @vapier @mheinzler @ngie-eign @jicama @maflcko @manxorist @ojwb @ggraham @peti). Still, I would prefer a solution to be merged rather quickly, as this is still completely preventing me from enabling C++23 with Autoconf in my project.

### DIFF
--- a/m4/ax_cxx_compile_stdcxx.m4
+++ b/m4/ax_cxx_compile_stdcxx.m4
@@ -44,7 +44,7 @@
 #   and this notice are preserved.  This file is offered as-is, without any
 #   warranty.
 
-#serial 24
+#serial 25
 
 dnl  This macro is based on the code from the AX_CXX_COMPILE_STDCXX_11 macro
 dnl  (serial version number 13).
@@ -167,35 +167,25 @@ m4_define([_AX_CXX_COMPILE_STDCXX_testbody_11],
 dnl  Test body for checking C++14 support
 
 m4_define([_AX_CXX_COMPILE_STDCXX_testbody_14],
-  [_AX_CXX_COMPILE_STDCXX_testbody_new_in_11
-   _AX_CXX_COMPILE_STDCXX_testbody_new_in_14]
+  [_AX_CXX_COMPILE_STDCXX_testbody_new_in_14]
 )
 
 dnl  Test body for checking C++17 support
 
 m4_define([_AX_CXX_COMPILE_STDCXX_testbody_17],
-  [_AX_CXX_COMPILE_STDCXX_testbody_new_in_11
-   _AX_CXX_COMPILE_STDCXX_testbody_new_in_14
-   _AX_CXX_COMPILE_STDCXX_testbody_new_in_17]
+  [_AX_CXX_COMPILE_STDCXX_testbody_new_in_17]
 )
 
 dnl  Test body for checking C++20 support
 
 m4_define([_AX_CXX_COMPILE_STDCXX_testbody_20],
-  [_AX_CXX_COMPILE_STDCXX_testbody_new_in_11
-   _AX_CXX_COMPILE_STDCXX_testbody_new_in_14
-   _AX_CXX_COMPILE_STDCXX_testbody_new_in_17
-   _AX_CXX_COMPILE_STDCXX_testbody_new_in_20]
+  [_AX_CXX_COMPILE_STDCXX_testbody_new_in_20]
 )
 
 dnl  Test body for checking C++23 support
 
 m4_define([_AX_CXX_COMPILE_STDCXX_testbody_23],
-  _AX_CXX_COMPILE_STDCXX_testbody_new_in_11
-  _AX_CXX_COMPILE_STDCXX_testbody_new_in_14
-  _AX_CXX_COMPILE_STDCXX_testbody_new_in_17
-  _AX_CXX_COMPILE_STDCXX_testbody_new_in_20
-  _AX_CXX_COMPILE_STDCXX_testbody_new_in_23
+  [_AX_CXX_COMPILE_STDCXX_testbody_new_in_23]
 )
 
 


ojwb:
Hmm, I've been using the version of the macro in my PR in what seems like essentially the same way you show and it seems to work for me:

```
AX_CXX_COMPILE_STDCXX([23], [], [optional])
if test "$HAVE_CXX23" = 0 ; then
  AX_CXX_COMPILE_STDCXX([20], [], [optional])
  if test "$HAVE_CXX20" = 0 ; then
    AX_CXX_COMPILE_STDCXX([17])
  fi
fi
```

Checking CI logs, this successfully detects compilers with C++23, C++20 and C++17 support in different jobs.

> However, as C++ standards are generally supersets of one another (and in cases where they are not, the user should expect that features which are removed in newer standards are actually not available when choosing such a standard version), only checking for features and/or the monotonically increasing __cplusplus value of the precise standard version requested, should be enough and yield identical results in all cases.

Not "all cases" - e.g. this patch would change behaviour of the macro if a compiler defines `__cplusplus` to the value for C++20 but actually has no (or buggy) support for something the macro currently tests for C++17 - currently it would not detect C++20 support for such a compiler, but with the patch it would.

I'd hope it's rare that a compiler sets `__cplusplus` to too high a value, but if we're going to not just trust its value then the current approach seems sensible.

I can take a look at what's going on.

@manxorist What compiler and version are you testing with?

manxorist:
> Hmm, I've been using the version of the macro in my PR in what seems like essentially the same way you show and it seems to work for me:
> 
> ```
> AX_CXX_COMPILE_STDCXX([23], [], [optional])
> if test "$HAVE_CXX23" = 0 ; then
>   AX_CXX_COMPILE_STDCXX([20], [], [optional])
>   if test "$HAVE_CXX20" = 0 ; then
>     AX_CXX_COMPILE_STDCXX([17])
>   fi
> fi
> ```

That's weird, so this really needs further investigation.

> > However, as C++ standards are generally supersets of one another (and in cases where they are not, the user should expect that features which are removed in newer standards are actually not available when choosing such a standard version), only checking for features and/or the monotonically increasing __cplusplus value of the precise standard version requested, should be enough and yield identical results in all cases.
> 
> Not "all cases" - e.g. this patch would change behaviour of the macro if a compiler defines `__cplusplus` to the value for C++20 but actually has no (or buggy) support for something the macro currently tests for C++17 - currently it would not detect C++20 support for such a compiler, but with the patch it would.

I would expect that to be rather rare. I guess it could happen on platforms with newer compiler but older library (which is a weird configuration in any case, but I think this can happen with Clang and libc++ on Android or Apple systems).

> I can take a look at what's going on.

Any help would be appreciated!

> @manxorist What compiler and version are you testing with?

This is on Ubuntu 22.04 (on WSL2), GCC 11.4.0, Autoconf 2.71. I can test other systems/versions tomorrow.

ojwb:
Thanks, I'm having a poke at it.

I notice there's some extra quoting added on top of my recently merged PR but the test source it compiles seems to be generated OK so that doesn't seem like the cause.

manxorist:
One confusing data point:
```
AX_CXX_COMPILE_STDCXX(20, [noext], [optional])
AS_IF([test "x$HAVE_CXX20" != "x1"],
 [
  AX_CXX_COMPILE_STDCXX(17, [noext], [mandatory])
 ],[]
)
```
works with a C++17 compiler (Debian 11, GCC 10.2.1, Autoconf 2.69)
```
checking whether g++ supports C++20 features with -std=c++20... no
checking whether g++ supports C++20 features with +std=c++20... no
checking whether g++ supports C++20 features with -h std=c++20... no
checking whether g++ supports C++20 features with -std:c++20... no
configure: No compiler with C++20 support was found
checking whether g++ supports C++17 features with -std=c++17... yes
```

On that system, the full C++23/C++20/C++17 cascade fails in exactly the same way as on the Ubuntu 22.04 system, though.

I cannot really make any sense out of that :/.

ojwb:
I think it's due to missing quoting in your input - this seems to work better:

```
AX_CXX_COMPILE_STDCXX(23, [], [optional])
AS_IF([test "x$HAVE_CXX23" != "x1"],
 [
 AX_CXX_COMPILE_STDCXX(20, [noext], [optional])
 AS_IF([test "x$HAVE_CXX20" != "x1"],
  [
   AX_CXX_COMPILE_STDCXX(17, [noext], [mandatory])
  ],[]
 )
 ]
)
```

Comparing the generated `configure` script with your version vs this, a lot of `[` and `]` are missing - just one example:

```
<       [maybe_unused] auto unused = f1();
---
>       [[maybe_unused]] auto unused = f1();
```

Not sure if the macros should cope with such underquoted where they are used - this might indicate underquoting in the macro too.

That also explains why your cut down version is OK.

manxorist:
> I think it's due to missing quoting in your input - this seems to work better:

Thank you very much @ojwb . I had somehow obviously missed that, even after staring 100 times at it. :)

Works for me now.

ojwb:
Actually, I think there's a bug in the macro too - I'll submit a fix shortly.

ojwb:
Hmm, so 5b0d43b767bf3f800f97892905b6d1ba50150f1a added quoting to the definitions of `_AX_CXX_COMPILE_STDCXX_testbody_11`, etc but `_AX_CXX_COMPILE_STDCXX_testbody_23` was missed (I suspect my patch adding it hadn't been merged when the patch for adding the quoting was submitted).

However adding that quoting doesn't seem to actually make any difference to the generated `configure` script, at least not for the example in the initial comment here.

If we're going to have that quoting it should be consistent between these macros so I will submit a patch, but I'll poke this a bit more first.

ojwb:
I haven't found a situation in which the quoting my PR adds actually makes a difference, but it does seem more correct to quote here and it really should be done consistently for all the _AX_CXX_COMPILE_STDCXX_testbody_* definitions.

I also tested underquoting the autoconf-provided macro `AC_PROG_CXX`, comparing the configure code generated for:

```
AC_PROG_CXX
```

```
AS_IF([true], AC_PROG_CXX)
```

and

```
AS_IF([true], [AC_PROG_CXX])
```

The middle one generates broken code, so it seems OK that an underquoted `AX_CXX_COMPILE_STDCXX` does too.

PR #289, opened 2023-09-26
Merge my 'autoupdate' branch

cooljeanius:
This isn't actually ready to merge yet (see cooljeanius/autoconf-archive#1), but I thought I'd open this as a draft anyways just so I can have something to look at while I fix the merge conflicts...

PR #310, opened 2024-10-16
Silence lint warnings on CI workflows

cooljeanius:
This is just the CI workflows portion of #289 that can be merged independently from the rest of it; I'm leaving out the parts that depend on the added targets or other changes to the build system made in that PR.

### DIFF
--- a/.github/workflows/autoconf-archive.yml
+++ b/.github/workflows/autoconf-archive.yml
@@ -1,22 +1,23 @@
+---
 name: cooljeanius/autoconf-archive
 on:
   push:
     branches:
-    - "**/*"
+      - "**/*"
   pull_request:
 jobs:
   test:
     runs-on: ubuntu-latest
     steps:
-    - name: checkout
-      uses: actions/checkout@v3.5.0
-      with:
-        fetch-depth: 50000
-    - run: sudo apt-get update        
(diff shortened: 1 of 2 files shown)


.github/workflows/autoconf-archive.yml
```diff
@@ -1,22 +1,23 @@
+---
```
alerque, line 1: What linter is throwing a warning about not having an opening YAML separator in GH workflow files? This just isn't a thing.

alerque (changes requested):

.github/workflows/autoconf-archive.yml
cooljeanius: `yamllint` is

alerque: This is a document separator and only relevant in the case where multiple YAML documents may be included in a single file. Such is not the case for GH workflow definitions and it is relatively rare to do so. Not invalid per se but not recommended by GH and typically not used because it is optional and could never be meaningful in this context.

cooljeanius: OK, I disabled the warning instead in [3c4cecd](https://github.com/autoconf-archive/autoconf-archive/pull/310/commits/3c4cecdea0c0a00d861252e82e6f4b3919017313).

PR #314, opened 2024-10-20
Create ax_check_awk__ooo_escapes.m4

mpsuzuki:
Hi, I want to propose a macro to detect "\123" style escape on AWK.

AX_CHECK_AWK__OOO_ESCAPES is similar to an existing macro AX_CHECK_AWK__X_ESCAPES,
but check the availability of triple octal escape. This escape syntax was introduced on Research
Unix V7 Addenda, thus, awk on the earliest V7, 32V, and the descendants forked from 32V
(like 2.9BSD, 3BSD) did not support it.


rurban:
Cannot you merge these two, and force push again? I think the first bad one blocks it

mpsuzuki:
...now, do I have any unfinished homework?

PR #318, opened 2024-11-09
Revert "Add LuaJIT support to AX_PROG_LUA macro"

peti:
Reverts autoconf-archive/autoconf-archive#288.

See https://github.com/autoconf-archive/autoconf-archive/pull/288#discussion_r1807567838 for further details.


alerque:
I think there is now a proper way forward in #313 so reverting should not be necessary.

thesamesam:
ping -- @peti, could you please merge either this or #313 and cut a new release?

PR #320, opened 2024-11-22
Add lua require macro and combinator for existing lua rock check

alerque:
As a follow up to #319 this adds much more robust functionality to searching for Lua rocks or modules. See commit messages for details on why these macros are useful. I've used a two step check like this for years on a couple projects but never cleaned up the macros for submission before.


alerque:
Changed to draft mode only because this builds on the commit in #319 and that PR should be merged first, then separately the 3 commits here.

PR #323, opened 2024-12-03
Don't embed build timestamp for AX_AM_MACROS*.

jas4711:
Hi!

I noticed that libtasn1 source tarballs are not reproducible because AX_AM_MACROS* embed timestamps into some *.m4 files.

Would you consider removing that timestamping, as per the attached patch?

I believe adding build timestamps into generated files like this is considered poor practice, and no other autoconf/automake/libtool/gnulib etc appear to do so.

/Simon

### DIFF
--- a/m4/ax_am_macros.m4
+++ b/m4/ax_am_macros.m4
@@ -24,7 +24,7 @@
 #   and this notice are preserved. This file is offered as-is, without any
 #   warranty.
 
-#serial 11
+#serial 12
 
 AC_DEFUN([AX_AM_MACROS],
 [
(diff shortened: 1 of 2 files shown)


alerque:
Definitely agree, macros like this should not be jinxing reproducible builds for no real gain. At the very least such a string injection would need to honor `SOURCE_DATE_EPOCH`, but I don't see the point of implementing that here because the default should be leaning towards reproducible builds anyway and in this case that output would be less valuable than no output at all. So not injecting the date at all is the best approach.

alerque (approved):

PR #326, opened 2025-01-20
ax_python.m4: add new python version

Tropicao:
Projects depending on ax_python detection may fail on systems with latest python version (3.13)

Add this new version to the list of searched versions


cpackham:
Looks like I've duplicated this on https://savannah.gnu.org/patch/index.php?10529

One thing you might want to include from my version is the handling of the distutils deprecation.

PR #329, opened 2025-06-08
AX_LIB_READLINE: remove a spurious line from the example usage code

rrthomas:
It looks like some project-specific code crept into the example usage code: there's no global called `cmdline` in readline, so setting a variable of that name seems to be otiose (not to mention possibly buggy!).


petk:
Thanks.

PR #337, opened 2026-03-27
clang doesn't support attr `malloc(func,index)` and `AX_GCC_FUNC_ATTRIBUTE` and `AX_GCC_MALLOC_CALL` don't check for it

dsalt:
This tripped me up recently when adding some checks to another project, so naturally I set about fixing it. Here's the result: tests which take account of clang only supporting `__attribute__((malloc))` and not gcc's extended form which specifies the `free` function and the argument index.

If support is present:
* `AX_GCC_FUNC_ATTRIBUTE([malloc_args])` will define `HAVE_FUNC_ATTRIBUTE_MALLOC_ARGS`=`1`;
* `AX_GCC_MALLOC_CALL` will define `GCC_MALLOC_CALL_FREE`=`__attribute__((__malloc__(func, index)))`.


dsalt:
Author: me, derived from existing code in the modified files as needed.

Licence: same as the files' existing licences (following future licensing changes at need).

PR #341, opened 2026-05-17
AX_PTHREAD: Fix pthread detection on Solaris 11.4

rorth:
`AX_PTHREAD` fails to detect pthread support on Solaris 11.4.  This was originally noticed with the copy of `ax_pthread.m4` in the GCC source tree (serial 24), but is still present in the current version:

	build: Fix ax_pthread.m4 on Solaris 11.4
	https://gcc.gnu.org/pipermail/gcc-patches/2026-May/716850.html

It turns out the problem is that `AX_PTHREAD` expects `_REENTRANT` to be defined on Solaris.  However, as described in

```
        https://gcc.gnu.org/gcc-16/porting_to.html#solaris
```
and
```
	Goodbye (And Good Riddance) to -mt -and -D_REENTRANT
	https://www.linker-aliens.org/blogs/ali/entry/mtreentrant/
```

referenced from there, Solaris 11.4 removed references to `_REENTRANT` from the Solaris system headers.  Given that, GCC 16 no longers defines it with `-pthread`.

This started when `libpthread` was folded into `libc` in Solaris 11.

However, `AX_PTHREAD` expects `_REENTRANT` to be defined in multithreaded programs on Solaris, which is no longer true.  To detect the situation, this patch checks for XPG7 support which was introduced together with those changes in Solaris 11.4.

However, the `_REENTRANT` check is dubious at best.  If a feature isn't detected, this may be because the corresponding feature test macros isn't defined.  However, the reverse is not true: a feature way well be available be default without any feature test macro.

The fix was tested in the GCC tree as described above, and manually on Solaris 11.3 where _REENTRANT is still referenced and needed.


cooljeanius (approved):

m4/ax_pthread.m4
petk:
```suggestion
#serial 32
```

